PR #247, opened 2014-10-30
Extension members are returned as members of current modules

### LINKED ISSUE #234: Type extensions aren't returned in the list of types of a module

dungpa:
Close #234.


7sharp9:
This only deals with `intrinsic` type extensions right?

dungpa:
I guess not. I think Intrinsic type extensions are exposed as normal members.

7sharp9:
Yeah they are, just checking :-)

Annoyingly types and methods with `[<Extension>]` are not considered to be type extensions :-(

PR #282, opened 2014-12-05
Update OnEvaluation event to have FSharpSymbolUse information available

7sharp9:
This also simplifies the API as FSharpSymbolUse contains the display
context and range.  This allows for a better accuracy by using pattern
matching on the symbols.

### DIFF
--- a/src/fsharp/fsi/fsi.fs
+++ b/src/fsharp/fsi/fsi.fs
@@ -149,12 +149,11 @@ type internal FsiValuePrinterMode =
     | PrintExpr 
     | PrintDecl
 
-type EvaluationEventArgs(name : string, displaycontext : FSharpDisplayContext, range : range, fsivalue : FsiValue) =
+type EvaluationEventArgs(name : string, fsivalue : FsiValue, symbolUse : FSharpSymbolUse) =
     inherit EventArgs()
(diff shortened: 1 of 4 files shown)


7sharp9:
Ok, added simple test.

7sharp9:
@dsyme ^

7sharp9:
Ping @anyone, is this ok or?   

I would like to get this merged so I don't have to use a branch , and instead use the nuget :-)

src/fsharp/fsi/fsi.fs
```diff
@@ -1120,23 +1119,23 @@ type internal FsiDynamicCompiler
 
         let filteredByVal =
             allNames
-            |> List.choose (function
```
dsyme: THis code is still kind of strange (as it was in the originakl version). It reports the top-level names defined, e.g. "x" in "let x = 1", but doesn't report any modules or types defined.

That's not related to this PR but the feature should eventually be adjusted to do a more systematic job on reporting freshly defined items.

7sharp9: Sure, the filter could be removed and to return non val items.  As far as I remember all the global top levels are here too like bigint etc

PR #290, opened 2015-01-09
update assembly info

duckmatt:


forki:
I think this should actuall be done automatically during release. See how we do this in paket: https://github.com/fsprojects/Paket/blob/master/build.fsx#L277

duckmatt:
The file is updated on build using the version from the release notes, it would be good to have something a task like in paket to commit to keep the file up to date on here

forki:
it's already in https://github.com/fsprojects/ProjectScaffold/blob/master/build.template#L298 so it would be awesome if you could steal it from there.

PR #296, opened 2015-02-10
Add Overloads to the symbols signature so it is publicly visible

7sharp9:
This should of been visible in the original commit but was missed unfortunatly.

This also adds the matchParameter number argument so either all
overloads, or overloads matching the current symbols parameter count
can be returned.

### DIFF
--- a/src/fsharp/vs/Symbols.fs
+++ b/src/fsharp/vs/Symbols.fs
@@ -1034,19 +1034,20 @@ and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) =
         | Some v -> Some v.Range
         | None -> base.DeclarationLocation 
 
-    member x.Overloads matchParamterNumber=
+    member x.Overloads matchParameterNumber =
         checkIsResolved()
         match d with
         | M m ->
             match item with
             | Item.MethodGroup (_name, methodInfos) ->
                 let methods =
-                    if matchParamterNumber then
+                    if matchParameterNumber then
                         methodInfos
                         |> List.filter (fun methodInfo -> not (methodInfo.NumArgs = m.NumArgs) )
                     else methodInfos
                 methods
                 |> List.map (fun mi -> FSharpMemberOrFunctionOrValue(cenv, M mi, item))
+                |> makeReadOnlyCollection
                 |> Some
             | _ -> None
         | _ -> None
--- a/src/fsharp/vs/Symbols.fs
+++ b/src/fsharp/vs/Symbols.fs
@@ -1034,14 +1034,18 @@ and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) =
         | Some v -> Some v.Range
         | None -> base.DeclarationLocation 
 
-    member x.Overloads =
+    member x.Overloads matchParamterNumber=
         checkIsResolved()
         match d with
         | M m ->
             match item with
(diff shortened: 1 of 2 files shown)


src/fsharp/vs/Symbols.fs
```diff
@@ -1034,14 +1034,18 @@ and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) =
         | Some v -> Some v.Range
         | None -> base.DeclarationLocation 
 
-    member x.Overloads =
+    member x.Overloads matchParamterNumber=
```
OkayX6: typo: matchParam**e**terNumber + missing space before the '='

src/fsharp/vs/Symbols.fsi
```diff
@@ -682,6 +682,10 @@ and [<Class>] FSharpMemberOrFunctionOrValue =
 
     member CurriedParameterGroups : IList<IList<FSharpParameter>>
 
+    /// Gets the overloads for the current method
+    /// matchParamterNumber indicates whether to filter the overloads to match the number of parameters in the current symbol
```
OkayX6: Same typo in comment :)

src/fsharp/vs/Symbols.fsi
```diff
@@ -682,6 +682,10 @@ and [<Class>] FSharpMemberOrFunctionOrValue =
...
+    member Overloads : bool -> FSharpMemberOrFunctionOrValue list option
```
OkayX6: In terms of consistency, shouldn't you return an ```IList<>``` or something like this? See the method above which returns an ```IList<...>```.

7sharp9: Others elsewhere return a std list, I don't mind IList if it's the preference.

OkayX6: As you wish, I'm just seeing 38 ```IList<>``` vs 3 ```list``` :)

src/fsharp/vs/Symbols.fsi
7sharp9: Late for dinner unsaved file commit :-)

src/fsharp/vs/Symbols.fs
```diff
@@ -1034,15 +1034,20 @@ and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) =
...
         | M m ->
             match item with
-            | Item.MethodGroup (_name, methodInfos) -> 
-                methodInfos
-                |> List.filter (fun methodInfo -> not (methodInfo.NumArgs = m.NumArgs) )
+            | Item.MethodGroup (_name, methodInfos) ->
+                let methods =
+                    if matchParameterNumber then
+                        methodInfos
+                        |> List.filter (fun methodInfo -> not (methodInfo.NumArgs = m.NumArgs) )
```
dsyme: I'm not quite sure what this check will do when there are optional and named parameters?

dsyme:
Hey Dave, could you add some tests, and some documentation to the tutorials under docs?  We need to always be doing both for all new API sureface area

src/fsharp/vs/Symbols.fs
7sharp9: Im happy to just return an unfiltered list, the filtering is then up to the consumer.

7sharp9:
I can this one is pretty low priority for me at the moment, although it seems to complete a missing part of the API in my opinion anyway :-)

PR #297, opened 2015-02-11
Added GetMethodsAsSymbols

7sharp9:
This unifies tooltip/completion/parameter completion so each one can be
queried as symbols.  The symbol information is used by the F# language
binding to display extra information and render tooltips and
completions semantically.

### DIFF
--- a/src/fsharp/vs/service.fs
+++ b/src/fsharp/vs/service.fs
@@ -1164,6 +1164,39 @@ type TypeCheckInfo
             (fun msg -> 
                 FSharpMethodGroup(msg,[| |]))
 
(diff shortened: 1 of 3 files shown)


dsyme:
Hi @7sharp9  - It looks good - could you add some tests?

7sharp9:
Yep

7sharp9:
@dsyme Ok, added test for this similar to `GetMethods` test in `EditorTests.fs`

7sharp9:
Not sure whats up with Travis, tests seem fine with AppVeyor though.

PR #298, opened 2015-02-13
Added curried parameter groups for nested functions

7sharp9:
This works much the same way as GetMethods in that only the types are involved in the function are used.  See the comment in service.fs line 170:

```
// ValReprInfo = None i.e. in let bindings defined in types or in local functions
// in this case use old approach and return only information about types
```

In this instance we use `stripFunTy` to produce the function argument types, and then in the for loop we use if `isTupleTy` and `tryDestTupleTy` if the argument is a tuple.  

This produces the same `CurriedParameterGroups` as if the function was in a module binding i.e.

```fsharp
let test one two three =
   one + two + three
```
This would be:
``` 
IList[IList[parameter1]
      IList[parameter2]
      IList[parameter3]]
```

```fsharp
let test (one, two, three) =
   one + two + three
```
This would be:
```
IList[IList[parameter1;parameter2;parameter3]]
```

### DIFF
--- a/src/fsharp/vs/Symbols.fs
+++ b/src/fsharp/vs/Symbols.fs
@@ -1407,15 +1407,8 @@ and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) =
                         then tryDestTupleTy cenv.g typ
                         else [typ]
                     yield
-                      [for typ in allArguments do
-                          match typ with
-                          | TType_var tp ->
-                              let isParamArrayArg = HasFSharpAttribute cenv.g cenv.g.attrib_ParamArrayAttribute tp.Attribs
-                              let isOutArg = HasFSharpAttribute cenv.g cenv.g.attrib_OutAttribute tp.Attribs && isByrefTy cenv.g typ
-                              let isOptionalArg = HasFSharpAttribute cenv.g cenv.g.attrib_OptionalArgumentAttribute tp.Attribs
-                              let argInfo : ArgReprInfo = { Name=None; Attribs= [] }
-                              yield FSharpParameter(cenv,  tp.AsType, argInfo, x.DeclarationLocationOpt, isParamArrayArg, isOutArg, isOptionalArg)
-                          | _ -> () ]
+                      allArguments
+                      |> List.map (fun arg -> FSharpParameter(cenv,  arg, { Name=None; Attribs= [] }, x.DeclarationLocationOpt, false, false, false))
                       |> makeReadOnlyCollection ]
                 |> makeReadOnlyCollection
             else
--- a/src/fsharp/vs/Symbols.fs
+++ b/src/fsharp/vs/Symbols.fs
@@ -1411,8 +1411,7 @@ and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) =
                       |> List.map (fun arg -> FSharpParameter(cenv,  arg, { Name=None; Attribs= [] }, x.DeclarationLocationOpt, false, false, false))
                       |> makeReadOnlyCollection ]
                 |> makeReadOnlyCollection
-            else
-                failwith "not a module let binding or member"
+            else makeReadOnlyCollection []
         | Some (ValReprInfo(_typars,curriedArgInfos,_retInfo)) -> 
             let tau = v.TauType
             let argtysl,_ = GetTopTauTypeInFSharpForm cenv.g curriedArgInfos tau range0
--- a/src/fsharp/vs/Symbols.fs
+++ b/src/fsharp/vs/Symbols.fs
@@ -1397,7 +1397,29 @@ and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) =
 
         | V v -> 
         match v.ValReprInfo with 
-        | None -> failwith "not a module let binding or member"
+        | None ->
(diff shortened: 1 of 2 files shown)


7sharp9:
@dsyme I added tests for this one now :-)

7sharp9:
Looks like the error is a spurious timeout etc.

7sharp9:
Does this look good to merge now?

src/fsharp/vs/Symbols.fs
```diff
@@ -1397,7 +1397,29 @@ and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) =
...
+                              let isParamArrayArg = HasFSharpAttribute cenv.g cenv.g.attrib_ParamArrayAttribute tp.Attribs
+                              let isOutArg = HasFSharpAttribute cenv.g cenv.g.attrib_OutAttribute tp.Attribs && isByrefTy cenv.g typ
+                              let isOptionalArg = HasFSharpAttribute cenv.g cenv.g.attrib_OptionalArgumentAttribute tp.Attribs
+                              let argInfo : ArgReprInfo = { Name=None; Attribs= [] }
+                              yield FSharpParameter(cenv,  tp.AsType, argInfo, x.DeclarationLocationOpt, isParamArrayArg, isOutArg, isOptionalArg)
+                          | _ -> () ]
+                      |> makeReadOnlyCollection ]
+                |> makeReadOnlyCollection
+            else
+                failwith "not a module let binding or member"
```
dsyme: This case would return empty

src/fsharp/vs/Symbols.fs
```diff
@@ -1397,7 +1397,29 @@ and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) =
...
+                let typeArguments, _typ = stripFunTy cenv.g tau
+                [ for typ in typeArguments do
+                    let allArguments =
+                        if isTupleTy cenv.g typ
+                        then tryDestTupleTy cenv.g typ
+                        else [typ]
+                    yield
+                      [for typ in allArguments do
+                          match typ with
+                          | TType_var tp ->
```
dsyme: I'm not sure what this is doing. It seems to be saying "if the argument's type is a generic type parameter T, then look at the attributes of the generic type parameter to work out if the argument is optional etc."  But that doesn't make sense.

I think isOptionalArg, isOutArg and isParamArrayArg should all just be false and you shouldn't be doing this match at all.

src/fsharp/vs/Symbols.fs
```diff
@@ -1397,7 +1397,29 @@ and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) =
...
+                              let isParamArrayArg = HasFSharpAttribute cenv.g cenv.g.attrib_ParamArrayAttribute tp.Attribs
+                              let isOutArg = HasFSharpAttribute cenv.g cenv.g.attrib_OutAttribute tp.Attribs && isByrefTy cenv.g typ
+                              let isOptionalArg = HasFSharpAttribute cenv.g cenv.g.attrib_OptionalArgumentAttribute tp.Attribs
+                              let argInfo : ArgReprInfo = { Name=None; Attribs= [] }
+                              yield FSharpParameter(cenv,  tp.AsType, argInfo, x.DeclarationLocationOpt, isParamArrayArg, isOutArg, isOptionalArg)
```
dsyme: "tp.AsType" can just be "typ"

tests/service/ProjectAnalysisTests.fs
```diff
@@ -4325,3 +4325,63 @@ let ``Test project34 should report correct accessibility for System.Data.Listene
...
+module Project35 = 
+    open System.IO
+
+    let fileName1 = Path.ChangeExtension(Path.GetTempFileName(), ".fs")
+    let base2 = Path.GetTempFileName()
+    let dllName = Path.ChangeExtension(base2, ".dll")
+    let projFileName = Path.ChangeExtension(base2, ".fsproj")
+    let fileSource1 = """
+type Test =
+    let curriedFunction (one:int) (two:float) (three:string) =
```
dsyme: Does  ValReprInfo = None for these declarations today?  From the look of what you have above, I was expecting the test case to be a function defined in an expression:

    let Test() =
         let curriedFunction (one:int) (two:float) (three:string) = ...

not a function defined in a class.

7sharp9: let bindings in a type have no `ValReprInfo` either, thats why Ive always moaned about `CurrendParameterGroups`

src/fsharp/vs/Symbols.fs
7sharp9: I kept the fail with as thats how unmatched results were represented before.  Empty would be better than a fail though.

src/fsharp/vs/Symbols.fs
7sharp9: The match also serves to get the `typar` `tp` which is the second param of `FSharpParameter`.  The attributes can be removed as the spec doesn't support them does it?  I wasn't positive so...

7sharp9:
@dsyme I think Ive addressed all the points.

7sharp9:
Anything else to do here?

haf:
+1 to this PR, see https://bugzilla.xamarin.com/show_bug.cgi?id=27452

PR #308, opened 2015-02-27
fixed: no FSharpSymbolUse for single case union type

vasily-kirichenko:


dsyme:
Hi @vasily-kirichenko , could you add a test case for this as well?  Thanks

vasily-kirichenko:
@dsyme done.

PR #322, opened 2015-03-26
Added Symbol Property for FSharpType

7sharp9:
Currently only TType's of type TType_app are converted into a symbol.
I can add further conversions if anyone can give me guidance how they would convert, preferable with a snippet of code to add to the included unit tests.

With the current implementation its possible to retrieve an FSharpSymbols BaseType property and use the Symbol property to do further queries with the symbol API.

I see this as a precursor to adding IDE commands like: "goto base", "find implementation of"

### DIFF
--- a/src/fsharp/vs/Symbols.fs
+++ b/src/fsharp/vs/Symbols.fs
@@ -1688,6 +1688,32 @@ and FSharpType(cenv, typ:TType) =
        protect <| fun () -> 
         "type " + NicePrint.stringOfTy (DisplayEnv.Empty(cenv.g)) typ 
 
+    member x.Symbol =
+       match typ with
+       /// Indicates the type is a universal type, only used for types of values and members 
+       | TType_forall (typars, tType) ->
+          None
+       /// Indicates the type is build from a named type and a number of type arguments
+       | TType_app (tyconRef, typeInst) ->
+          let item = Item.Types(tyconRef.DisplayName, [typ])
+          let symbol = FSharpSymbol.Create(cenv, item)
+          Some symbol
+       /// Indicates the type is a tuple type. elementTypes must be of length 2 or greater.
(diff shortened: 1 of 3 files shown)


7sharp9:
@dsyme If its not easy to figure out symbols for the other TTypes then I can adjust this so that instead of FsharpType having a Symbol property, I can add a `BaseTypeSymbol` to `FsharpEntity`

I think `GetSuperTypeOfType` always only returns something if the TTYpe is a `TType_app` otherwise the result is an object.

I did take a look through the code to see how I might construct something from a `TType_fun` function type but I could not see the relation ship between that and the `Item` type which ends up becoming a Symbol.

Any ideas on any of this?

src/fsharp/vs/Symbols.fs
```diff
@@ -1688,6 +1688,32 @@ and FSharpType(cenv, typ:TType) =
        protect <| fun () -> 
         "type " + NicePrint.stringOfTy (DisplayEnv.Empty(cenv.g)) typ 
 
+    member x.Symbol =
```
dsyme: I think it would be better to factor this as an "AsSymbol()" method on FSharpEntity.  Then you use fSharpType.TypeDefinition.AsSymbol().

Could you make that change?

7sharp9: Yep, I can do that on Monday, to be fair I couldn't see an easy way to deal with all the TTypes from an FSharpType.

7sharp9: @dsyme Actually, isn't FSharpEntity already a symbol?

dsyme: Well, yes, now that you mention it! So I don't think you need this PR, you just use fSharpType.TypeDefinition guarded by HasTypeDefinition

PR #332, opened 2015-04-15
Fix output location of referenced projects

rneatherway:
Prior to this the referencing project file's path would be concatenated
to the relative path to the referenced project's output directory. If
either one of these was the same across the two projects (usually the
latter) this bug was masked.

Also added a test.

Also disabled PCL tests on Linux where PCL is normally not available and
these tests fail. Is there a better way to do this?

This fixes a bug encountered by @7sharp9.

### DIFF
--- a/src/fsharp/vs/service.fs
+++ b/src/fsharp/vs/service.fs
@@ -2367,9 +2367,9 @@ type FSharpProjectFileInfo (fsprojFileName:string, ?properties, ?enableLogging) 
             if String.IsNullOrWhiteSpace v then None
             else Some v
 
-        let outdir p = mkAbsoluteOpt directory (getProp p "OutDir")
+        let outdir = mkAbsoluteOpt directory (getProp project "OutDir")
         let outFileOpt p =
-            match outdir p with 
+            match outdir with
             | None -> None
             | Some d -> mkAbsoluteOpt d (getProp p "TargetFileName")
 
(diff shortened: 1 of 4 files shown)


dsyme:
We only want to disable those tests for Linux, not Mac I think. Or is there some way to prove for the existence of PCL binaries on Linux?

rneatherway:
Yes, confusingly though Mac has its own platform id separate from Unix. I
think PCL on Linux is rare enough not to worry about right now. I haven't
actually checked on Mac that the tests still run, that would be good to
know.

7sharp9:
I can test tomorrow morning, it would be good to get another nuget out as the current one is pretty broken. 

Cheers
D.

dsyme:
Hi @rneatherway - could you take a look at the AppVeyor test failure? It looks related:

    https://ci.appveyor.com/project/fsgit/fsharp-compiler-service/build/0.0.88-a1504151635

7sharp9:
@dsyme @rneatherway Tests all work for me on a mac.

7sharp9:
@rneatherway From the look of AppVeyor the bug looks to still exist in the new msbuild api version:

"C:\projects\fsharp-compiler-service\tests\service\data\DifferingOutputDir\Dir2\OutputDir2\Test1.dll"
vs
"C:\projects\fsharp-compiler-service\tests\service\data\DifferingOutputDir\Dir1\OutputDir1\Test1.dll"

rneatherway:
Dave do the PCL tests run on your Mac? Just want to double check they
aren't skipped. I'll take a look at the appveyor results soon.

7sharp9:
@rneatherway Yep, they do run.

rneatherway:
OK, the New API was using the correct output path of the subproject, but I have changed it so that it is consistent with the Old API. They both now use the copy of the assembly that ends up in the output folder of the main project.

dsyme:
Ah, got it, thanks!

PR #340, opened 2015-05-18
Fixes #339, releases memory from projectStringCache at the end of the function

### LINKED ISSUE #339: GetProjectOptionsFromProjectFile seems to retain a large portion of memory.
Is it possible to get this released earlier?

Even with a moderate project fileIm seeing 180MB in a ProjectStringCache which appears to be allocated on this call path.

This is also mentioned here:  https://connect.microsoft.com/VisualStudio/feedback/details/816008/memory-leak-in-projectstringcache-with-net-4-5

Heres a call stack showing creation:
screen shot 2015-05-15 at 22 08 00

7sharp9:
After loading projects projectStringCache is now no longer retaining memory:
screen shot 2015-05-18 at 11 06 37

### DIFF
--- a/src/fsharp/vs/service.fs
+++ b/src/fsharp/vs/service.fs
@@ -2450,7 +2450,7 @@ type FSharpProjectFileInfo (fsprojFileName:string, ?properties, ?enableLogging) 
                    | Some (Some output) -> yield output
                    | _ -> ()
               ]
-
+        engine.UnloadAllProjects()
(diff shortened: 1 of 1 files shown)


dsyme:
Did this fix not work? thx

7sharp9:
It feed the memory but I had trouble determining if the results came back properly.  It appeared that it was returning empty results.  The overall function is not thread safe so I ended up with a lock around it.  FCS could wrap it in a reactor op, not had time to do that though.

Issue #343, opened 2015-05-29
GetProjectOptionsFromProjectFile` doesn't work for projects that reference a pcl library


7sharp9:
Take an executable that references Sqllite, a sample project can be obtained here: https://bugzilla.xamarin.com/attachment.cgi?id=9100

I used a gist for the output from FCS and xbuild as they clog things up here:
https://gist.github.com/7sharp9/f3728050a8a87309d1da

7sharp9:
Essentially all the implicit pcl references are missing so the type checker gives you invalid warnings:

![screen shot 2015-05-29 at 16 57 23](https://cloud.githubusercontent.com/assets/588746/7886736/f21a0b96-0623-11e5-968f-3489ed990ca5.png)

But everything compiles just fine.

7sharp9:
@rneatherway @dsyme Any ideas on this would be appreciated

In debug it looks like: `DesignTimeFacadeAssemblies` contains the 49 missing references:

![screen shot 2015-05-29 at 18 35 24](https://cloud.githubusercontent.com/assets/588746/7888581/7e7e4202-0631-11e5-9f95-da4f27f4a964.png)

7sharp9:
But should the tasks we pass into msbuild be automagically producing the correct list of references?

7sharp9:
Just a note, unfortunately adding assembly redirects doesn't magically fix this issue its still needs resolving, not that I thought it would fix this one.

Theres a distinct lack of detail on how portable stuff resolves references, I guess we just need to reverse engineer the msbuild targets to mach sure we are calling all the correct tasks and collecting the correct properties.  

@rneatherway Do you have any notes on this while writing this feature?

rneatherway:
I notice in your output there is: `Target named 'ImplicitlyExpandTargetFramework' not found in the project.`

We add that here, and this is what should automagically produce the correct list: https://github.com/fsharp/FSharp.Compiler.Service/blob/master/src/fsharp/vs/service.fs#L2361

This seems likely to be the culprit.

rneatherway:
That was a comment I wrote but forgot to post on Friday! I have to admit that I know very little about the portable profiles. If you could add your repro project as a test in a branch I will have a poke at it.

7sharp9:
`ImplicitlyExpandTargetFramework` is not in the fsproj being analysed but in the pcl that is referenced by that project.

rneatherway:
[...]

Issue #294, opened 2015-02-05
Project cracking code fails on Websharper projects


rneatherway:
First reported at https://github.com/fsharp/fsharpbinding/issues/945

The Websharper project uses a custom task. The project cracking code fails with an invalid cast exception when trying to run it. This can be reproduced using the project given at the fsharpbinding issue above and the small script below:

```fsharp
#r @"./FSharp.Compiler.Service.0.0.82/lib/net45/FSharp.Compiler.Service.dll"

open System
open System.Collections.Generic
open Microsoft.FSharp.Compiler.SourceCodeServices

let projectFile  = __SOURCE_DIRECTORY__ + @"/sleep_log.fsproj"
let p = FSharpProjectFileInfo.Parse(projectFile, enableLogging=true)
Console.WriteLine(p.LogOutput)
```

Error:

```
InvalidCastException, ITask: Microsoft.Build.Framework.ITask, Microsoft.Build.Framework, Version=12.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a Task type: IntelliFactory.WebSharper.MSBuild.WebSharperTask, IntelliFactory.WebSharper.MSBuild, Version=2.5.0.0, Culture=neutral, PublicKeyToken=dcd983dec8f76a71
Error initializing task WebSharperTask: Cannot cast from source type to destination type.
Error initializing task WebSharperTask: System.InvalidCastException: Cannot cast from source type to destination type.
```

Probably some issue with the `ToolsVersion` as this is project is `4.0`. However, I tried recompiling FCS to always use `4.0` as a test and this fails with an error we saw before:

```
InvalidCastException, ITask: Microsoft.Build.Framework.ITask, Microsoft.Build.Framework, Version=12.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a Task type: Microsoft.Build.Tasks.GetFrameworkPath, Microsoft.Build.Tasks.v4.0, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a
Error initializing task GetFrameworkPath: Cannot cast from source type to destination type.
Error initializing task GetFrameworkPath: System.InvalidCastException: Cannot cast from source type to destination type.
```

It's possible that we have already bound to the ToolsVersion 12 MSBuild assemblies, not sure exactly what the problem is.

rneatherway:
This is no longer a problem, although I don't know why.

PR #345, opened 2015-06-01
Add fix for #343 Use ResolveReferences task

7sharp9:
This task seems to stably return the reference list now.
Added a test using an executable that references a pcl assembly


rneatherway:
We should enable all the PCL tests on linux now that that is supported.

7sharp9:
Do you want to do that as a separate PR or?

rneatherway:
I think it's best here so we can see the tests passing across the board.

rneatherway:
Otherwise your new test isn't exercised on Unix.

7sharp9:
Seems some tests are failing, not sure if the correct tasks are in place for all situations, might have to drop this completly if I cant get it working tomorrow.

dsyme:
Yes, two failing: 

```
9191) Test Failure : FSharp.Compiler.Service.Tests.ProjectOptionsTests.Project file parsing -- 2nd level references
2920     Expected: "Found 'FSharp.Core.dll'"
2921Actual: "Failed to find 'FSharp.Core.dll'"


29312) Test Failure : FSharp.Compiler.Service.Tests.ProjectOptionsTests.Project file parsing -- Exe with a PCL reference
2932     Expected: set
2933  ["FSharp.Core.dll"; "SQLite.Net.Platform.Generic.dll";
2934   "SQLite.Net.Platform.Win32.dll"; "SQLite.Net.dll";
2935   "System.Collections.Concurrent.dll"; "System.Collections.dll";
2936   "System.ComponentModel.Annotations.dll";
2937   "System.ComponentModel.EventBasedAsync.dll"; "System.ComponentModel.dll"; ...]
2938Actual: set
2939  ["FSharp.Core.dll"; "System.Core.dll"; "System.Numerics.dll"; "System.dll";
2940   "mscorlib.dll"]

```

7sharp9:
Ha switching to `ResolveReferences` fixes the references pcl issue but breaks all other use cases, it seems that the evaluated item `ReferencePath` is not empty but the evaluated item `Reference` exists although the references are not path resolved just fully qualified names.

rneatherway:
Can you explain a bit more? Almost all the tests pass so things don't seem too bad! Your new test project should probably be under `data` with the other non-canonical tests, and I think it will have some difficulty passing because it references assemblies from the non-existent `packages` directory (I think SQLite.Net comes from NuGet). Perhaps they can just be removed?

rneatherway:
I do see that things are worse on Windows though.

rneatherway:
Should have posted this here:

2nd level references test fails on Linux. Output includes:

```
InvalidCastException, ITask: Microsoft.Build.Framework.ITask, Microsoft.Build.Framework, Version=12.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a Task type: Microsoft.Build.Tasks.Message, Microsoft.Build.Tasks.v4.0, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a
Error initializing task Message: Cannot cast from source type to destination type.
Error initializing task Message: System.InvalidCastException: Cannot cast from source type to destination type.
  at Microsoft.Build.BuildEngine.BuildTask.InitializeTask () [0x00000] in <filename unknown>:0 
```

Classic old friend

rneatherway:
I've pushed what I tried here, which seems decent, although your new test doesn't pass. As I said, the SQLite dlls are not there so it isn't that surprising.

7sharp9:
Having pcl working is the most important think for me, it's unshippable without that.  :-(

rneatherway:
Have you tried it with the SQLite dlls present? Perhaps it works fine. All the other PCL tests work OK.

7sharp9:
Yes Ive tried, we are no further forward its still exactly the same issue, no implicit pcl references are present.

7sharp9:
@rneatherway I just corrected the package folder, although it makes no difference even if you restore the package.

7sharp9:
@rneatherway The only way I can figure at the moment is evaluating the names evaluation group `_DesignTimeFacadeAssemblies` and if it exists return those assemblies too.

rneatherway:
Hmm. I don't understand that because we are running `ImplicitlyExpandDesignTimeFacades`. What is different when compared to this project and all the others where the PCL works? What is the diff between the log outputs?

7sharp9:
Heres a gist of the 259 vs sqllite test:
https://gist.github.com/7sharp9/9f1466ae40b86919a8f1

7sharp9:
It makes no difference, mine has the package restored

7sharp9:
With the sqllite one there is an extra property `_DesigntimeFacadeAssemblies` which is not present in the 259 test.

7sharp9:
The only reference I can find is here:
https://github.com/Microsoft/msbuild/blob/82177a50da735cc0443ac10fa490d69368403d71/src/XMakeTasks/Microsoft.NETFramework.CurrentVersion.targets#L99-L102

rneatherway:
Hey did you notice this at the end of the sqlite one:

> Target named 'ImplicitlyExpandTargetFramework' not found in the project.

7sharp9:
Yep, that target doesn't exist as its a property of the pcl reference rather than the project under test

rneatherway:
What do you mean? That target is found in Microsoft.Common.targets. If I build your sqlite project with xbuild then it runs that target.

7sharp9:
Im guessing its not looking there

rneatherway:
The previous line, though, is 

> Done building target "ResolveAssemblyReferences" in project "/Users/dave/code/FCS/tests/projects/sqlite-net-spike/sqlite-net-spike.fsproj" ("/Library/Frameworks/Mono.framework/Versions/4.0.1/lib/mono/4.5/Microsoft.Common.targets").

I think that perhaps the issue is that it doesn't run `ImplicitlyExpandTargetFramework` and then as a result it doesn't run `ImplicitlyExpandDesignTimeFacades` either because it is after it in the list of targets.

7sharp9:
By the ways its not just the sqllite test project that behaves this way and project that references a pcl is missing the references.

rneatherway:
Understood. When you first submitted this PR did it work, but it just broke other things?

7sharp9:
Yeah, everything else broke but referenced pcl projects worked.

I had a quick look in monodevelop to see how the overrided GetReferences method worked in C#, it looks like the framework references are appended here:
https://github.com/mono/monodevelop/blob/b10a2f2b566a33ec333d5b9e5fa3d1c1e16829cf/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/PortableDotNetProjectFlavor.cs#L79-L85

7sharp9:
Just to clarify a single `ResolveReferences" works perfectly:

screen shot 2015-06-02 at 18 26 56

rneatherway:
Thanks, very interesting. Is that back with the v12 MSBuild or v4?

I found that in the case where the test fails it is instead pulling in the assemblies from `mono/xbuild-frameworks/.NETPortable/v4.5` rather than `mono/4.5/Facades/`, which includes more assemblies. The difference is the flag settings:

```
~/dev/other/FSharp.Compiler.Service $ monodis --assembly /usr/lib/mono/4.5/Facades/System.Xml.XmlSerializer.dll 
Assembly Table
Name:          System.Xml.XmlSerializer
Hash Algoritm: 0x00008004
Version:       4.0.0.0
Flags:         0x00000001
PublicKey:     BlobPtr (0x00000001)
```
```
~/dev/other/FSharp.Compiler.Service $ monodis --assembly /usr/lib/mono/xbuild-frameworks/.NETPortable/v4.5/System.Xml.XmlSerializer.dll 
Assembly Table
Name:          System.Xml.XmlSerializer
Hash Algoritm: 0x00008004
Version:       4.0.0.0
Flags:         0x00000071
PublicKey:     BlobPtr (0x0000028c)
```

7sharp9:
Thats with this current PR with the single target `ResolveReferences`.

Maybe we should work backwards, with this PR pcl projects resolve correctly as do projects that reference a pcl, the thing that doesnt work is the `Project file parsing -- 2nd level references` test which only returns one reference.

7sharp9:
If I enable logging the results look curious:
https://gist.github.com/7sharp9/04743db202e079f943c0

rneatherway:
That's really odd, but I think I see the issue. For some reason it actually goes off to build Test1 (referenced by Test2, the main project file), and this then fails because the source files aren't present, and why should they be! So perhaps this would work in a normal situation where the referenced project actually compiles, but it still seems a bit strict, and also slow! We don't want to compile all the referenced projects I don't think.

rneatherway:
Right, I found this: https://github.com/mono/mono/blob/master/mcs/tools/xbuild/data/14.0/Microsoft.Common.targets#L345

>		

Which then calls the MSBuild task. Because it hasn't been built, it then builds it. I don't think we want to do this. We fake it later anyway by loading the referenced projects and finding their `OutputFile`.

Happily, I found that swapping the order of the multi-task option seems to work. Try:

```
project.Build([| "ResolveAssemblyReferences";
                         "ImplicitlyExpandDesignTimeFacades";
                         "ImplicitlyExpandTargetFramework" |])  |> ignore
```

so that it doesn't matter if `ImplicitlyExpandTargetFramework` isn't present.

rneatherway:
Oh god. That link I sent shows that `ResolveProjectReferences` only builds the referenced projects if `$(BuildingInsideVisualStudio) = 'true'`. Do you know how to set a variable like that?

rneatherway:
Thanks MS for posting open-sourcing Visual F# tools: https://github.com/Microsoft/visualfsharp/blob/fsharp4/vsintegration/src/vs/FsPkgs/FSharp.Project/Common.Source.CSharp/Project/ProjectNode.cs#L3432

rneatherway:
OK, it appears that our victory is complete. We just have to use an undocumented variable to pretend to be Visual Studio. Obvious really...

@7sharp9 please can you add your android project as a test so that we can make sure we don't break that in the future.

I'm not clear on whether we should be using MSBuild 4 or 12 really. The other scenarios that I'm aware of (xbuild, and now the Visual F# Tools) both use conditional compilation to prepare different versions compiled against different versions of the MSBuild assemblies. Perhaps we should do the same and pass the choice onto consumers of FCS.

rneatherway:
This should also fix #342 and #294.

rneatherway:
Ah right, so without v12 it fails on Windows. Mono is more robust. The tests pass on my Linux machine.

7sharp9:
@rneatherway I can try adding the android project, although Android will not be installed on the CI systems will it?

rneatherway:
Oh it needs some extra packages, sorry I didn't realise. Probably not the way to go then.

rneatherway:
Because I had to revert back to MSBuild 12 so that it works on Windows, this will not fix the android issue anymore, at least not without binding redirects.

dsyme:
I've lost track of this conversation - could someone summarize the state of play? :) Thanks!

rneatherway:
Hi Don, well it took a lot of effort, but I think it is a unquestioned improvement. The headline is that we are using the target `ResolveReferences` rather than the old `ResolveAssemblyReferences; ImplicitlyExpandTargetFramework`. This should run all the same tasks that the build would run, rather than hackishly adding extra targets ourselves. One such is `ResolveProjectReferences`, which prevents us having to recurse on the projects manually. However, it then actually tries to build the referenced projects, because that is how `Microsoft.Common.targets` works *unless* you set the property "BuildingInsideVisualStudio" to "true". So I've done that. This should be much more robust.

However, there are two outstanding issues:

1. The sqlite test (project referencing a PCL-using DLL) doesn't pass on Windows. To debug that I need to see the `LogOutput` from FCS compared to `msbuild /v:diag` on Windows. I think it may be a ToolsVersion issue, see '2'.

2. The ToolsVersion and MSBuild version assembly issue.
    a) On Windows, using MSBuild v4 causes many tests to fail, presumably the 'New API' is not well supported if at all in those assemblies.
    b) Generally, using MSBuild v12 causes Websharper/Android/custom Tasks that inherit from MSBuild v4 Task (e.g. #342 and #294) to fail with an InvalidCastException. We could try, although I haven't yet, using a binding redirect e.g. on FSharp.AutoComplete as Xbuild v12 does. However, xbuild doesn't use a binding redirect for `Microsoft.Build.Tasks.vXX.dll` and I don't see how that could work because the assemblies have different names. So I'm thinking that perhaps we are not setting the engine/project `ToolsVersion` correctly.

I'm concerned that we are possibly setting ourselves an impossible task in having a single version of FCS that works for all projects and installed versions of MSBuild. Both xbuild and the Visual F# VS integration use conditional compilation and thus assume that a particular instance of MSBuild is present. Perhaps @tpetricek's plan to bundle a particular version, which aligns with this model, is actually wise.

I have noticed that the Visual F# VS integration invokes the target `Compile` and only falls back to `ResolveReferences`: https://github.com/Microsoft/visualfsharp/blob/fsharp4/vsintegration/src/Salsa/salsa.fs#L223

7sharp9:
@rneatherway Does running the `Compile` target on windows make it succeed?

rneatherway:
No, it doesn't make any difference. The log output from FCS is much less useful on Windows, see here: https://gist.github.com/rneatherway/5c8028fa0b84a40e3fa9

For some reason it just doesn't give the same level of detail as msbuild does, while on Mono that works fine.

7sharp9:
Hmm, well they're both building with v12 due to:
```
Project file contains ToolsVersion="4.0". This toolset may be unknown or missing, in which case you may be able to resolve this by installing the appropriate version of MSBuild, or the build may have been forced to a particular ToolsVersion for policy reasons. Treating the project as if it had ToolsVersion="12.0". For more information, please see http://go.microsoft.com/fwlink/?LinkId=293424.
```

In XS with C# the target reference assemblies are added to the returned references if the project context is pcl anyway.  This stuff is pretty difficult to resolve nicely it seems.

rneatherway:
Yeah, I noticed that. I was somewhat heartened because at least they are both doing the same thing. It's hard to see what the problem is exactly because of the logger not being verbose enough. It might be fruitful to try the same logger that they use in Visual F# tools: https://github.com/Microsoft/visualfsharp/blob/fsharp4/vsintegration/src/Salsa/salsa.fs#L219

If it's possible to determine that the project reference is PCL somehow then could we just hackily add the target reference assemblies?

7sharp9:
I think you would have to enumerate the target references and project references and detect it there, or alternatively there will probably be a property thats set thats states that this is the case.  The `_DesignTimeFacadeAssemblies` has the reference list so its just doing the detection thats the difficult bit.

rneatherway:
Is property populated on Windows?

src/fsharp/vs/service.fs
```diff
@@ -2442,16 +2434,12 @@ type FSharpProjectFileInfo (fsprojFileName:string, ?properties, ?enableLogging)
...
-                   yield i.EvaluatedInclude
-                 for fsproj in projectReferences do
-                   match (try let p' = projectInstanceFromFullPath fsproj
-                              Some (outFileOpt p') with _ -> None) with
-                   | Some (Some output) -> yield output
-                   | _ -> ()
-              ]
+              [ for i in project.GetItems("ReferencePath") do
+                  yield i.EvaluatedInclude
+                for i in project.GetItems("ChildProjectReferences") do
```
dsyme: Later, we return "projectReferences" and make it available as the "ProjectReferences" property.  Should we instead be returning the information from "ChildProjectReferences"?

tests/service/ProjectOptionsTests.fs
```diff
@@ -199,7 +172,9 @@ let ``Project file parsing -- reference project output file``() =
   p.OutputFile
   |> should equal (Some expectedOutputPath)
 
-  p.References |> should contain (normalizePath (__SOURCE_DIRECTORY__ + @"/data/DifferingOutputDir/Dir2/OutputDir2/Test1.dll"))
+  p.References
+  |> List.map (fun (s: string) -> s.Replace("//", "/"))
+  |> should contain (normalizePath (__SOURCE_DIRECTORY__ + @"/data/DifferingOutputDir/Dir1/OutputDir1/Test1.dll"))
```
dsyme: Why has this test result changed? Is this expected from the fixes being made?

src/fsharp/vs/service.fs
rneatherway: "projectReferences" contains paths to the fsproj files referenced, while "ChildProjectReferences" contains the output assemblies.

tests/service/ProjectOptionsTests.fs
rneatherway: Yes, this is expected. Previously if Test2.fsproj referenced Test1.fsproj then we would use Test2/bin/Debug/Test1.dll (the CopyLocal version), but ResolveReferences returns Test1/bin/Debug/Test1.dll (the originally produced version). I think it doesn't really matter which, and the new behaviour is consistent with visualfsharp.

dsyme: OK, thanks

src/fsharp/vs/service.fs
dsyme: OK, thanks

dsyme:
@rneatherway  Re [problem 2 above](https://github.com/fsharp/FSharp.Compiler.Service/pull/345#issuecomment-108495620) - as far as I can see from my comment here https://github.com/fsharp/FSharp.Compiler.Service/issues/342#issuecomment-106627742, the cast failure is when casting from type

    Task type: Xamarin.Android.Tasks.ResolveSdks, Xamarin.Android.Build.Tasks, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null

to 

    ITask: Microsoft.Build.Framework.ITask, Microsoft.Build.Framework, Version=12.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a 

Note that the target interface type being cast to is in Microsoft.Build.Framework (which looks like it will need a binding redirect from 4.0.0.0 --> 12.0.0.0 for things to work), not Microsoft.Build.Tasks.vXX.dll (which doesn't need a binding redirect).  

It seems to me that adding the 4.0.0.0 --> 12.0.0.0 binding redirect for Microsoft.Build.Framework  would solve this problem in the cases where we can assume MSBuild 12.0.0.0 on the target machine. However I also notice Tomas' comment here: https://github.com/fsharp/FSharp.AutoComplete/issues/11, which says we can't particularly assume either MSBuild 12.0.0.0 or 14.0.0.0 on the target machine for tools like FsAutocomplete.exe (though it might be reasonable to expect one of the two)  In that case I've no real idea how one is meant to deliver tools that depend minimally on the MSBuild v12.0.0.0 API, except by bundling the MSBuild 12.0.0.0 DLLs.with your application.  Perhaps we should ask someone on the MSBuild team for guidance about what they expect people to do for such tools.

Could you add a failing test for problem (2)?

rneatherway:
Ah, my mistake. I assumed that as it was a `Task` it was in the Tasks dll. You're right, the binding redirect should be fine there.

I tried to construct a test for problem (2), but the WebSharper issue no longer occurs and I don't have access to Android projects. @7sharp9 how could we add a test for that?

I'm strongly leaning towards bundling a version of MSBuild with FCS as being the correct resolution for this issue, but it would be great to ask somebody in the know.

As for the windows issue, with massive thanks again to MS for the recent open-sourcing spree, I have found a suitable logger to subsitute for my `BasicStringLogger` (https://github.com/Microsoft/msbuild/blob/master/src/XMakeBuildEngine/Logging/SerialConsoleLogger.cs) so I should be able to get some more debug output for that tonight.

7sharp9:
@rneatherway I dont know, Android projects will require the android msbuild tasks to be present on the cI server:

```
  <Import Project="$(MSBuildExtensionsPath)\Xamarin\Android\Xamarin.Android.FSharp.targets" />
```

dsyme:
You could make a new task DLL, e.g. by copying and modifying [this stuff](https://github.com/fsharp/fsharp/tree/master/src/fsharp/FSharp.Build)?  Or [this DLL](https://github.com/fsharp/fsharp/blob/master/lib/bootstrap/4.0/FSharp.Build.dll) or [this DLL](https://github.com/fsprojects/FsLexYacc/blob/master/lkg/bin/FsLexYacc.Build.Tasks.dll) are MSBuild task DLLs with a 4.0 dependency?  I presume trying to load that as a custom task will fail?

rneatherway:
Should be OK to just modify the paths of the targets file and custom task assembly then put them in the project directory. If you do that then we probably don't want to commit the assemblies to this repo but we could at least test whether it is working on an ad-hoc basis. I'm not aware of any other custom tasks that exhibit this behaviour now that Websharper works. If you know of any we could use that instead.

7sharp9:
The other thing you could do to harden the testing is to invoke type checking on the project and assert that there are no Errors present.

rneatherway:
Well I tried using a `ConsoleLogger`, but unfortunately for some reason it doesn't give and more information really, just:

```
Target "ResolveAssemblyReferences" in file "C:\Program Files (x86)\MSBuild\12.0\bin\amd64\Microsoft.Common.CurrentVersion.targets":
  Using "ResolveAssemblyReference" task from assembly "Microsoft.Build.Tasks.v12.0, Version=12.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a".
  Task "ResolveAssemblyReference"
  Done executing task "ResolveAssemblyReference".
Done building target "ResolveAssemblyReferences" in project "sqlite-net-spike.fsproj".
```

But not what and how it happened. By contrast, running msbuild.exe from the command line gives much information:

```
Target "ResolveAssemblyReferences: (TargetId:12)" in file "C:\Program Files (x86)\MSBuild\12.0\bin\amd64\Microsoft.Common.CurrentVersion.targets" from project "D:\dev\FSharp.Compiler.Service\tests\service\data\sqlite-net-spike\sqlite-net-spike.fsproj" (target "ResolveReferences" depends on it):
Using "ResolveAssemblyReference" task from assembly "Microsoft.Build.Tasks.v12.0, Version=12.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a".
Task "ResolveAssemblyReference" (TaskId:7)
  Task Parameter:
      Assemblies=
          mscorlib
          FSharp.Core
          System
          System.Numerics
          SQLite.Net.Platform.Win32
                  HintPath=..\..\..\..\packages\SQLite.Net-PCL.3.0.5\lib\net4\SQLite.Net.Platform.Win32.dll
          SQLite.Net
                  HintPath=..\..\..\..\packages\SQLite.Net-PCL.3.0.5\lib\net40\SQLite.Net.dll
          SQLite.Net.Platform.Generic
...
<snip>
```

but I don't know the name of this logger to use it programmatically.

rneatherway:
OK... apparently there is a property called `ResolveAssemblyReferencesSilent` which disables logging for that task specifically. I'll have to try that, but it will be after the weekend.

```
<ResolveAssemblyReferencesSilent 
  Condition="'$(ResolveAssemblyReferencesSilent)' == '' and
  '$(TraceDesignTime)' != 'true' and
  '$(BuildingProject)' == 'false'">
    true
</ResolveAssemblyReferencesSilent>
<ResolveAssemblyReferencesSilent 
  Condition="'$(ResolveAssemblyReferencesSilent)' == ''">
    false
</ResolveAssemblyReferencesSilent>
```

7sharp9:
It would be good to get this one resolved so that project cracking is usable :-(

rneatherway:
I couldn't get it working without using the `Build` target. Of course this then actually builds things, which is slow and therefore bad. So I ripped off the sneaky HostCompilation trick from visualfsharp's salsa.fs, which skips the compilation step and keeps things fast. A console app with a proof of concept is at https://github.com/rneatherway/msbuildresolution, and is much easier to test than waiting 10 minutes for the FCS unit tests.

rneatherway:
This seems to work locally. I had to relax the condition on Dave's new PCL test because Windows uses a slightly different set of assemblies, but I think it still suffices to check that the PCL reference assemblies are being pulled in.

dsyme:
@rneatherway sounds like progress!

rneatherway:
It is my belief that this is ready to be merged now. I don't think it fixes #342 (android), but it would be nice to have confirmation from @7sharp9. I've come to also believe that this should be fixed by using an assembly binding redirect (and bundling MSBuild a la #338).

7sharp9:
A binding redirect would have to be applied to the originator process, unless this was invoked oop

rneatherway:
Yes I understand. I think this is consistent with xbuild, msbuild and Visual Studio.

7sharp9:
Ok, I was going over this issue with one of the guys here and he was of the opinion that this function should be done out of process otherwise unknown external dlls could be loaded into the monodevelop process.  We have a similar method in MonoDevelop which is running targets for MonoDevelop, all out of process for these reasons.

rneatherway:
Ah right, because of the custom tasks etc? That makes sense. Would that mean having an exe wrapper that just printed the results on standard out? Or is there a cleverer way of doing that in .NET?

Either way, I would be glad to merge this and do a release as it at least fixes the PCL referencing issue and generally makes things more robust. @dsyme are you OK with that?

7sharp9:
@rneatherway Did you make changes to both new and old project cracking?

rneatherway:
Yes, the old API now uses the `ResolveReferences` with the `BuildingInsideVisualStudio` property set. It's actually easier with the old API than the new.

7sharp9:
@rneatherway MonoDevelop does it like this:
https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.Formats.MSBuild/MSBuildProjectService.cs#L504-L608

And theres a reference resolution task already in existence too, I might have linked you to this a week back maybe.
https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.Formats.MSBuild/RemoteProjectBuilder.cs#L188-L216

rneatherway:
I see. I think they should be using `ResolveReferences` there with the property set like I did. They can probably avoid adding the PCL references afterwards manually that way.

I notice this comment:

```
//always start the remote process explicitly, even if it's using the current runtime and fx
//else it won't pick up the assembly redirects from the builder exe
```

is pretty telling. I suppose we could always use that method, but I'm pretty happy with where we are now.

7sharp9:
Having unknown references polluting the current runtime process is quite nasty too so its better to be an isolationist here I think.

dsyme:
@7sharp9 - it seems it would be best to implement the isolation via an appdomain or an external process. Can you do that as part of XS/MD addin implementation or do you think we should do it in FCS? 

@rneatherway - yes we can merge and re-release.

7sharp9:
@dsyme Well I would expect to be able to call an exposed method without having to wrap it in another process, it probably should be part of FCS.

Ive been told calling msbuild has to be done out of process, so I wont be able to use this functionality as it stands. :-(

dsyme:
@7sharp9 - well, you can use it, you just have to call it out of process :)  

I agree it would be good to do that automatically though, I suppose by including a fsservice.exe in the "tools" directory that can be invoked behind the scenes

7sharp9:
At the moment I don't have time to refactor to use out of process though, Ill have to do that before this release goes out, I suppose Android is broken until then too.

PR #348, opened 2015-06-09
Add compiler location as a static member of CompilerLocation

7sharp9:
BinFolderOfDefaultFSharpCompiler is copied and used in both fsautocomplete and xamarin-monodevelop-fsharp-addin this PR makes that property available to both without having to duplicate code.


rneatherway:
Thanks!

PR #349, opened 2015-06-13
Fix for styles loading

sergey-tihon:
I hope that this change should fix following errors:
image


rneatherway:
I meant to say, I tried to regenerate the docs yesterday after you released
the NuGet package (thanks @dsyme) and it failed because the RazorEngine
NuGet package has some dependency issues and no net40 version which the fsx
script depends on. Do you have a clone with a working version of the
package? Perhaps we should add a pinned fetch in build.cmd, or use the
newer project scaffold technique which doesn't seem to need Razor.

PR #359, opened 2015-07-01
switch from NuGet to Paket

ctaggart:


rneatherway:
Tested as working on Linux.

ctaggart:
Thank you @rneatherway for trying it out. :+1:

Issue #258, opened 2014-11-09
Running Compile tutorial code produces StackOverflowException in ImplicitLoadIfAllowed


adamkennedy:
I have been experimenting with dynamic assembly compilation and the results of my tests have been going really well, while I was launching the compiler from F# Interactive sessions.

However, when switching to a Visual Studio Command Line F# project with the FSharp.Compiler.Service 0.76 NuGet package (so I can move to building a production version) I am getting StackOverflowExceptions from infinite loops in the compiler service.

This occurs regardless of the code being compiled.

For example, creating a project with the Hosted Compiler example code results in the same StackOverflowException as my actual project code.

``` fsharp
// Learn more about F# at http://fsharp.net
// See the 'F# Tutorial' project for more help.

open Microsoft.FSharp.Compiler.SimpleSourceCodeServices
open System.IO

let scs = SimpleSourceCodeServices()

[<EntryPoint>]
let main argv = 
    let fn = Path.GetTempFileName()
    let fn2 = Path.ChangeExtension(fn, ".fs")
    let fn3 = Path.ChangeExtension(fn, ".dll")

    File.WriteAllText(fn2, """
    module M

    type C() = 
       member x.P = 1

    let x = 3 + 4
    """)

    let errors2, exitCode2, dynAssembly2 = 
        scs.CompileToDynamicAssembly([| "-o"; fn3; "-a"; fn2 |], execute=None)

    // return an integer exit code
    0
```

I have traced the infinite loop to the following set of stack entries.

```
FSharp.Compiler.Service.dll!Microsoft.FSharp.Compiler.Build.TcImports.RegisterAndImportReferencedAssemblies(Microsoft.FSharp.Core.FSharpOption<Microsoft.FSharp.Core.FSharpFunc<string,Microsoft.FSharp.Core.Unit>> displayPSTypeProviderSecurityDialogBlockingUI, Microsoft.FSharp.Collections.FSharpList<Microsoft.FSharp.Compiler.Build.AssemblyResolution> nms) Line 4408   F#
    FSharp.Compiler.Service.dll!Microsoft.FSharp.Compiler.Build.tryFile@4420(Microsoft.FSharp.Compiler.Build.TcImports tcImports, Microsoft.FSharp.Compiler.Range.range m, string speculativeFileName) Line 4425    F#
[...]

PR #362, opened 2015-07-02
Add noframework option in AST compiler methods

eiriktsarpalis:
I added this option to the AST compiler methods in order to work around [this issue](https://github.com/fsharp/FSharp.Compiler.Service/issues/258#issuecomment-62381514) in my [QuotationCompiler](https://github.com/eiriktsarpalis/QuotationCompiler/) library.

### DIFF
--- a/src/fsharp/fsc.fs
+++ b/src/fsharp/fsc.fs
@@ -1964,10 +1964,11 @@ let main1(tcGlobals,tcImports : TcImports,frameworkTcImports,generatedCcu,typedA
 
 
 // set up typecheck for given AST without parsing any command line parameters
-let main1OfAst (openBinariesInMemory, assemblyName, target, outfile, pdbFile, dllReferences, exiter, errorLoggerProvider: ErrorLoggerProvider, inputs) =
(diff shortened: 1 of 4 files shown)


7sharp9:
So this removes an implicit FSharp.Core reference?

eiriktsarpalis:
Yes, same as `--noframework` in `fsc.exe`.

7sharp9:
@dsyme Why is there no `Compile*` methods in normal `SourceCodeServices`?

I think that the API should be normalised that way, its annoying to have to create another type to Compile when you are already using `FsharpChecker` in SourceCodeServices which essentially the same API.

dsyme:
@7sharp9 - yes, it would be reasonable to bring the Compile methods into FSharpChecker. For starters that could then help ensure the same Reactor thread is used for both checking and compilation.

dsyme:
This fix is good enough considering it only affects the compile-from-AST methods.  

There is certainly duplication between the AST and compile-from-text paths which ideally should be resolved in the future, e.g. the command-line arguments governing the compilation (such as noframework) should probably come inn textual form, for uniformity with the rest of the compiler services.  But that would be a different commit.

So I'm ok to merge this.

eiriktsarpalis:
Thanks. Looking forward for a nuget package update!

PR #365, opened 2015-07-07
Exposes the DynamicAssembly so that it can be saved to disk by consumers.

voiceofwisdom:
I am working on using this to cache the compilation of FAKE's build script. This change lets you evaluate changes, and then save the in memory compiled assembly to disk using AssemblyBuilder.Save.

```f#
    let session = FsiEvaluationSession.Create(fsiConfig, commonOptions, stdin, outStream, errStream)
    session.EvalScript scriptPath
    let assemBuilder = session.DynamicAssembly :?> System.Reflection.Emit.AssemblyBuilder
    assemBuilder.Save("FSI-ASSEMBLY.dll")
```

### DIFF
--- a/src/fsharp/fsi/fsi.fs
+++ b/src/fsharp/fsi/fsi.fs
@@ -2532,6 +2532,8 @@ type FsiEvaluationSession (fsiConfig: FsiEvaluationSessionHostConfig, argv:strin
     member x.CurrentPartialAssemblySignature = 
         fsiDynamicCompiler.CurrentPartialAssemblySignature (fsiInteractionProcessor.CurrentState)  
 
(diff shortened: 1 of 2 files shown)


7sharp9:
Is there not an issue with optdata and sigdata missing from the assembly?

voiceofwisdom:
Surprisingly no, I was able to load the assembly via reflection and execute the main method. I did have to save it as "FSI-ASSEMBLY" though.

voiceofwisdom:
I think I didnt have any issue with those missing because I am only loading the assembly and executing it through reflection. If you were to attempt to reference it from a project there would probably be issues.

PR #370, opened 2015-07-08
Suggestion to bump version to 1.3.1.0 (3.1 indicates language version)

dsyme:
Suggestion to bump the FCS version

   1.3.1.x - latest for F# 3.1
   1.4.0.x - latest for F# 4.0

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,7 @@
+#### 1.3.1.0 -
+* simplified source indexing with new SourceLink
+* Add noframework option in AST compiler methods
+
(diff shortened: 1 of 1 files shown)


dsyme:
@rneatherway @forki - what do you think of this way to handle the multi-stream versioning?

rneatherway:
Seems very sensible to me. What should I do with FsAutoComplete and the editors? Just switch to 1.4.0.x at some point? Or perhaps release 2 versions of those for some time?

dsyme:
I think just move to 1.4.0.x when you gain confidence it does the job.

Issue #356, opened 2015-06-24
FSharpMemberOrFunctionOrValue.Overloads does not return overloads


7sharp9:
For example the following code:

``` fsharp
type JustOverloads() =
  member x.Overload(i:int) =
    i + 42
  member x.Overload(s:string) =
    int s + 42
```

The symbol `Overload` is a `MemberFunctionOrValue` but is assigned a `ValRef` type so the `Overloads` code doesnt ever return an item:

``` fsharp
    member x.Overloads matchParameterNumber =
        checkIsResolved()
        match d with
        | M m ->
            match item with
            | Item.MethodGroup (_name, methodInfos) ->
                let methods =
                    if matchParameterNumber then
                        methodInfos
                        |> List.filter (fun methodInfo -> not (methodInfo.NumArgs = m.NumArgs) )
                    else methodInfos
                methods
                |> List.map (fun mi -> FSharpMemberOrFunctionOrValue(cenv, M mi, item))
                |> makeReadOnlyCollection
                |> Some
            | _ -> None
        | _ -> None
```

Im not sure why the members are created as `ValRef` rather than `MethInfo` is it the case that only C# types with overloads are assigned that way?

alfonsogarciacaro:
I checked this with last build and now `Overloads()` returns an item, but it's always the same method without references to other overloads.

In above example, calling `[JustOverloads Entity].MembersFunctionsAndValues.[1].Overloads(false)` will give `Some(seq [Overload(i:int)])` but you don't get information about `Overload(s:string)`.

Moreover, if you call `Overloads(false)` on a `FSharpMemberOrFunctionOrValue` taken from a typed expression you still don't get any item.

PR #375, opened 2015-07-10
Implements FSharpMemberOrFunctionOrValue.IsConstructorThisValue, IsMemberThisValue & LiteralValue

alfonsogarciacaro:
I didn't implement IsBaseValue, because I couldn't find a case where you can get a `base` reference without using `BasicPatterns.BaseValue` active pattern (which doesn't include a `FSharpMemberOrFunctionOrValue` value).

Also adds some tests for these properties and for the `FSharpMemberOrFunctionOrValue.Overloads`. The latter doesn't pass at the moment, see #356.

Can you please review the PR? Cheers!

### DIFF
--- a/src/fsharp/vs/Symbols.fs
+++ b/src/fsharp/vs/Symbols.fs
@@ -118,7 +118,28 @@ module Impl =
         | FSharpOrArrayOrByrefOrTupleOrExnTypeMetadata -> 
             entity.Accessibility
 
-
+    let getLiteralValue = function
+        | Some lv  ->
+            match lv with
+            | Const.Bool    v -> Some(box v)
+            | Const.SByte   v -> Some(box v)
+            | Const.Byte    v -> Some(box v)
+            | Const.Int16   v -> Some(box v)
+            | Const.UInt16  v -> Some(box v)
+            | Const.Int32   v -> Some(box v)
+            | Const.UInt32  v -> Some(box v)
+            | Const.Int64   v -> Some(box v)
+            | Const.UInt64  v -> Some(box v)
+            | Const.IntPtr  v -> Some(box v)
+            | Const.UIntPtr v -> Some(box v)
+            | Const.Single  v -> Some(box v)
+            | Const.Double  v -> Some(box v)
+            | Const.Char    v -> Some(box v)
+            | Const.String  v -> Some(box v)
+            | Const.Decimal v -> Some(box v)
+            | Const.Unit
+            | Const.Zero      -> None
+        | None -> None
             
 
     type cenv(g:TcGlobals, thisCcu: CcuThunk , tcImports: TcImports) = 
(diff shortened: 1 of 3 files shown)


src/fsharp/vs/Symbols.fs
```diff
@@ -1491,18 +1492,33 @@ and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) =
...
     /// Is this the "x" in "type C() as x = ..."
-    member __.IsConstructorThisValue : bool
-
-    /// Is this the "x" in "member __.M = ..."
-    member __.IsMemberThisValue : bool
+    member __.IsConstructorThisValue =
+        if isUnresolved() then false else
+        match d with
+        | M _ | P _ | E _ -> false
+        | V v -> match v.BaseOrThisInfo with CtorThisVal -> true | _ -> false
```
dungpa: Could be shorted to `v.BaseOrThisInfo = CtorThisVal`.

src/fsharp/vs/Symbols.fs
```diff
@@ -1491,18 +1492,33 @@ and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) =
...
 
-    /// Is this a [<Literal>] value, and if so what value?
-    member __.LiteralValue : obj // may be null
+    /// Is this the "x" in "member x.M = ..."
+    member __.IsMemberThisValue =
+        if isUnresolved() then false else
+        match d with
+        | M _ | P _ | E _ -> false
+        | V v -> match v.BaseOrThisInfo with MemberThisVal -> true | _ -> false
```
dungpa: Could be shorted to `v.BaseOrThisInfo = MemberThisVal `.

tests/service/ProjectAnalysisTests.fs
```diff
@@ -4408,3 +4411,75 @@ let ``Test project35 CurriedParameterGroups should be available for nested funct
...
+    let options =  keepAssemblyContentsChecker.GetProjectOptionsFromCommandLineArgs (projFileName, args)
+    let wholeProjectResults =
+        keepAssemblyContentsChecker.ParseAndCheckProject(options)
+        |> Async.RunSynchronously
+    let declarations =
+        let checkedFile = wholeProjectResults.AssemblyContents.ImplementationFiles.[0]
+        match checkedFile.Declarations.[0] with
+        | FSharpImplementationFileDeclaration.Entity (_, subDecls) -> subDecls
+        | _ -> failwith "unexpected"
+    let getExpr exprIndex =
```
dungpa: I'm curious why you don't get all symbol uses and check their properties as per https://github.com/alfonsogarciacaro/FSharp.Compiler.Service/blob/dcfd9f816d807ff68fd120f6c406f65193b69f58/tests/service/ProjectAnalysisTests.fs#L4363-L4365. It seems complicated to walk the typed trees to assert simple properties.

tests/service/ProjectAnalysisTests.fs
```diff
@@ -4408,3 +4411,75 @@ let ``Test project35 CurriedParameterGroups should be available for nested funct
...
+    match Project36.getExpr 3 with
+    | BasicPatterns.Call(_,_,_,_,[BasicPatterns.Value s;_])
+        when not s.IsMemberThisValue && not s.IsConstructorThisValue -> ()
+    | _ -> failwith "val s in B.Overload(s) must not be MemberThis"
+
+    let project36Module = Project36.wholeProjectResults.AssemblySignature.Entities.[0]
+    let lit = project36Module.MembersFunctionsAndValues.[0]
+    let notLit = project36Module.MembersFunctionsAndValues.[1]
+    if lit.LiteralValue.IsNone || notLit.LiteralValue.IsSome then
+        failwith "val lit must be LiteralValue while notLit musn't"
```
dungpa: Please try to use NUnit assertions (e.g. shouldEqual, etc.) rather than raising ad-hoc exceptions.

tests/service/ProjectAnalysisTests.fs
```diff
@@ -4408,3 +4411,75 @@ let ``Test project35 CurriedParameterGroups should be available for nested funct
...
+
+[<Test>]
+let ``Test project36 FSharpMemberOrFunctionOrValue.Overloads(false)`` () =
```
dungpa: Do you plan to fix the bug related to `Overloads` in this PR? If not, this test doesn't belong to this PR.

dungpa:
Look good. It would be great to fix the failing test if it's meant to be a part of the changes.

alfonsogarciacaro:
Thanks a lot for your useful comments, @dungpa!

I've fixed the PR. You're right about the overloads (I had a look at the issue, but unfortunately my limited knowledge of the project didn't get me too far), so I deleted the test.

About walking the typed tree in the tests, I'm using the compiler services for FunScript where I'm accessing the symbols mostly through the AST and, as I've found some disagreements when you retrieve a symbol via the AST or something like `GetAllUsesOfAllSymbols` (as with `Overloads()`), I want to be sure the properties also work properly in the former case. I hope this is OK, but if it's not, tell me and I'll change the unit tests.

Cheers!

dungpa:
Great jobs. Thanks a lot.

I'm not aware that anyone uses typed tree API extensively. There are probably inconsistencies to iron out.

Issue #383, opened 2015-07-21
Wrong FSharp.Core resolution for F# 4.0


dungpa:
Tested with https://www.nuget.org/packages/FSharp.Compiler.Service/1.4.0-beta

FCS still uses FSharp.Core 4.3.0.0 as default. Consequently, any new functionality on FSharp.Core 4.4.0.0 doesn't work on scripts. This breaks all VFPT functionalities on scripts.

Here is a relevant log:

> GetScriptCheckerOptions: Creating for stand-alone file or script: 'C:\Users\dungpa\Documents\GitHub\Script1.fsx'
GetScriptCheckerOptions: ProjectFileName: C:\Users\dungpa\Documents\GitHub\Script1.fsx.fsproj, ProjectFileNames: [|"C:\Users\dungpa\Documents\GitHub\Script1.fsx"|], FSharpProjectOptions: [|"--noframework"; "--warn:3";
  "-r:C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.0\System.Numerics.dll";
  "-r:C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.0\mscorlib.dll";
  "-r:C:\Program Files (x86)\Reference Assemblies\Microsoft\FSharp\.NETFramework\v4.0\4.3.0.0\FSharp.Core.dll";
  "-r:C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.0\System.dll";
  "-r:C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.0\System.Xml.dll";
  "-r:C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.0\System.Runtime.Remoting.dll";
  "-r:C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.0\System.Runtime.Serialization.Formatters.Soap.dll";
  "-r:C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.0\System.Data.dll";
  "-r:C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.0\System.Drawing.dll";
  "-r:C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.0\System.Core.dll";
[...]

PR #384, opened 2015-07-22
fix 383

dsyme:
Tentative fix for #383 

The docs for FSharp.Core are also updated

### DIFF
--- a/docs/content/corelib.fsx
+++ b/docs/content/corelib.fsx
@@ -63,6 +63,23 @@ convert them to command-line arguments is to [crack an F# project file](http://f
 Alternatively you can compute SDK paths yourself, and some helpers to do this are in [the tests for FSharp.Compiler.Service.dll](https://github.com/fsharp/FSharp.Compiler.Service/blob/8a943dd3b545648690cb3bed652a469bdb6dd869/tests/service/Common.fs#L54).
 
 
+What about if I am processing a script or using ``GetCheckOptionsFromScriptRoot``
+-------------------------------------------------------------------------
+
+If you do _not_ explicitly reference an FSharp.Core.dll from an SDK location, or if you are processing a script
+using ``FsiEvaluationSession`` or ``GetCheckOptionsFromScriptRoot``, then an implicit reference to FSharp.Core will be made
+by the following choice:
+
+1. The version of FSharp.Core.dll statically referenced by the host assembly returned by ``System.Reflection.Assembly.GetEntryAssembly()``. 
+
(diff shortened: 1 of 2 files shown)


dungpa:
I thought about this again. It's unreasonable to expect cracking multiple versions of FSharp.Core inside FCS. I will do FSharp.Core cracking from VFPT as discussed.

I tested this PR on VFPT, it seems to work fine. 

Could you publish a non-beta NuGet package support F# 4.0? This branch is already quite stable and current FCS beta causes all consumers (Fantomas, FSharpLint.Core, FSharpVSPowerTools.Core) to be beta as well.

rneatherway:
@dungpa will you ship with one version of FCS (i.e. the F# 4.0 version) for installation into all VS versions? I'm wondering if I can do the same for FsAutoComplete.

dungpa:
> will you ship with one version of FCS (i.e. the F# 4.0 version) for installation into all VS versions? 

Yes, that's my plan. Based on VS versions, we will crack compiler arguments to provide suitable FSharp.Core versions for `GetCheckerOptionsFromScript`.

rneatherway:
OK, that makes sense. That VS context is useful, I guess I will just default to FSharp.Core 4.4.0.0.

7sharp9:
Can you post details of the fix for the other editors?

rneatherway:
@dungpa/@dsyme: just to clarify, is there any downside to using FSharp.Core 4.4.0.0 for scripts originally developed against older versions of FSharp.Core? Also, what would the effect be of deploying a version of FsAutoComplete on a machine without F# 4.0 installed? I assume it would work OK except for this issue with returning FSHarp.Core 4.4.0.0, which wouldn't exist in that case?

dsyme:
@rneatherway Don't forget the FSharp.Core used by the host process (if any) will be preferred. So the FSharp.Core used by FsAutoComplete.exe will be preferred.

If your host process doesn't reference FSharp.Core then the downside is that F# scripts won't typecheck correctly at all unless you have FSharp.Core 4.4.0.0 (and sigdata/optdata) available on your system.

rneatherway:
OK, so it sounds like it is completely safe for me to update to FCS 1.4.0.X, thanks!

As I'm about to also ask if @dungpa can link to his implementation, I do wonder if this could be a utility function in FCS after all. I assume we want to probe for the latest version of FSharp.Core available on the system and offer that?

dsyme:
@dungpa FCS 1.4.0.2 released as non-alpha

dungpa:
@dsyme Thanks.

Issue #363, opened 2015-07-05
Provide better API for FSharpMemberOrFunctionOrValue.


7sharp9:
It the moment you have to go through quite convoluted pattern matching or testing to determine if something as simple as whether a binding is a method call, constructor, event, or plain value.

I think there should be at least a way of finding out is something is a plain binding like

``` fsharp
module Test =
  let square x = x * x
  let twoSquared = square 2
```

If you retrieve a symbol for `twoSquared` you should be able to `twoSquaredMemberIOrVal.IsValue` and `twoSquaredMemberIOrVal.IsModuleFunction` for `square`

7sharp9:
I think having `Member` `Function` `Operator` `Constructor` and `Value` as valid subtypes would perhaps be a better option and slim down the number of properties on `FSharpMemberOrFunctionOrValue`.

I love adding tool functionality to tooling but working with the symbol API is not fun to be honest.

7sharp9:
closing old issue, fsharp.editing may address this I guess, eventually

PR #386, opened 2015-07-25
Created subtypes for FSharpMemberOrFunctionOrValue resolve #299

### LINKED ISSUE #299: How to represent Roslyn symbols from F# symbols
Im opening this to discuss how we can represent the roslyn symbols from our F# symbols.

In FCS we have the following symbols:

``` fsharp
FSharpMemberOrFunctionOrValue
FSharpUnionCase
FSharpEntity
FSharpField
FSharpGenericParameter
FSharpActivePatternCase
FSharpParameter
FSharpStaticParameter
```

Do we want to expand these so that they more closely match what would be available in Roslyn for easier interop.  We treat module and namespaces as `FSharpEntities` which I think matches the representation elsewhere in the API.

Roslyn has a lot more symbols.

``` csharp
IAliasSymbol
IAssemblySymbol
IEventSymbol
IFieldSymbol
ILabelSymbol
ILocalSymbol
IMethodSymbol
IModuleSymbol
INamespaceOrTypeSymbol
IParameterSymbol
IPreProcessorSymbol
IPropertySymbol
IRangeVariableSymbol
```

Those are just the core interfaces that inherit from `ISymbol`, for each of those interfaces there are numerous other subclasses.  i.e. INamespaceOrType`is implemented by`AssemblySymbol`which has subclasses`MissingAssemblySymbol`and`NonMissingAssemblySymbol`, which intern also have subclasses.

From a quick glance we could make FCS return finer grained symbol which would make matching the symbols easier, but there are other things like preprocessor symbols which are not available in FCS.  

My preference would be to expand the range of symbols as Ive always found the symbol API in FCS to be too coarse.

Thoughts?

jeroldhaas:
The symbol subtypes point to the main type for now since their properties are being used at the moment in many places to differentiate their type. This will allow backwards compatibility for the time being.

Additional features can be added as needed later.

### DIFF
--- a/src/fsharp/vs/Symbols.fs
+++ b/src/fsharp/vs/Symbols.fs
@@ -1006,8 +1006,6 @@ and FSharpModuleOrNamespaces = FSharpMemberOrFunctionOrValue
 and FSharpSetterArg = FSharpMemberOrFunctionOrValue
 and FSharpCustomBuilder = FSharpMemberOrFunctionOrValue
 and FSharpCustomOperation = FSharpMemberOrFunctionOrValue
-// TODO: duplicate type def - might be a custom CE keyword
-//and FSharpCustomOperation = FSharpMemberOrFunctionOrValue
 and FSharpTypeVar = FSharpMemberOrFunctionOrValue
 and FSharpActivePatternResult = FSharpMemberOrFunctionOrValue
 and FSharpArgName = FSharpMemberOrFunctionOrValue
--- a/src/fsharp/vs/Symbols.fs
+++ b/src/fsharp/vs/Symbols.fs
@@ -991,6 +991,31 @@ and FSharpMemberOrValData =
 
 and FSharpMemberOrVal = FSharpMemberOrFunctionOrValue
 
+// TODO: stubbed for now
+and FSharpValue = FSharpMemberOrFunctionOrValue
+and FSharpExnCase = FSharpMemberOrFunctionOrValue
+and FSharpRecdField = FSharpMemberOrFunctionOrValue
+and FSharpEvent = FSharpMemberOrFunctionOrValue
+and FSharpProperty = FSharpMemberOrFunctionOrValue
+and FSharpMethodGroup = FSharpMemberOrFunctionOrValue
+and FSharpCtorGroup = FSharpMemberOrFunctionOrValue
+and FSharpDelegateCtor = FSharpMemberOrFunctionOrValue
+and FSharpUnqualifiedType = FSharpMemberOrFunctionOrValue
+and FSharpTypes = FSharpMemberOrFunctionOrValue
+and FSharpModuleOrNamespaces = FSharpMemberOrFunctionOrValue
+and FSharpSetterArg = FSharpMemberOrFunctionOrValue
+and FSharpCustomBuilder = FSharpMemberOrFunctionOrValue
+and FSharpCustomOperation = FSharpMemberOrFunctionOrValue
+// TODO: duplicate type def - might be a custom CE keyword
+//and FSharpCustomOperation = FSharpMemberOrFunctionOrValue
+and FSharpTypeVar = FSharpMemberOrFunctionOrValue
+and FSharpActivePatternResult = FSharpMemberOrFunctionOrValue
+and FSharpArgName = FSharpMemberOrFunctionOrValue
+and FSharpImplicitOp = FSharpMemberOrFunctionOrValue
+and FSharpILField = FSharpMemberOrFunctionOrValue
+and FSharpFakeInterfaceCtor = FSharpMemberOrFunctionOrValue
+and FSharpNewDef = FSharpMemberOrFunctionOrValue
+
 and FSharpMemberFunctionOrValue =  FSharpMemberOrFunctionOrValue
 
 and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) = 


7sharp9:
You have made everything a `FSharpMemberOrFunctionOrValue` when not every Item becomes one of those.  I think the best first split it to work on the subtypes of `FSharpMemberOrFunctionOrValue` which will be namely `Method | Function | InnerFunction | Constructor | Operator`, `Property`, `Event`, `Value`.

I know I would like to have the first set as separate functions too as we have to reconstruct all of those subtypes in `xamarin-monodevelop-fsharp-addin` and VFPT also does this.

7sharp9:
See theres for the items that become `FSharpMemberOrFunctionOrValue`:
https://github.com/fsharp/FSharp.Compiler.Service/blob/master/src/fsharp/vs/Symbols.fs#L1850
https://github.com/fsharp/FSharp.Compiler.Service/blob/master/src/fsharp/vs/Symbols.fs#L1855
https://github.com/fsharp/FSharp.Compiler.Service/blob/master/src/fsharp/vs/Symbols.fs#L1858
https://github.com/fsharp/FSharp.Compiler.Service/blob/master/src/fsharp/vs/Symbols.fs#L1861
https://github.com/fsharp/FSharp.Compiler.Service/blob/master/src/fsharp/vs/Symbols.fs#L1864
https://github.com/fsharp/FSharp.Compiler.Service/blob/master/src/fsharp/vs/Symbols.fs#L1879
https://github.com/fsharp/FSharp.Compiler.Service/blob/master/src/fsharp/vs/Symbols.fs#L1882

`CustomOperation` and `CustomBuilder` are computation expression related for defined custom operations and keywords.

dsyme:
FWIW I'm not really in favour of creating a subtype hierarchy for FSharpMemberOfFunctionOrValue.  I don't really believe that any specific subtype hierarchy will be "right" in the face of the wide variety of factorizations possible for these types (e.g. should "Active Pattern" be a subtype?  I'd rather not go there).

I would instead be willing to consider well-thought out and well-considered set of active patterns (either partial or total) to make working with FSharpMemberOrFunctionOrValue a little easier.  Even then, it's a bit hard to get this right, but at least the use of any particular set of active patterns is entirely optional (and can sit comfortably beside an alternative set of active patterns which factor the space covered by FSharpMemberOrFunctionOrValue differently).

I just wanted to explain my position on this. (btw this doesn't resolve #299 which is a question about a Roslyn bridge - which would be a wonderful thing to have and a good set of active patterns could help establish that).

jeroldhaas:
My point in creating such an anemic PR first-off was to spur disscussion on how this should be properly implemented: I'd like to have the specification on how these are to be handled done as well as possible before I make attempts on how these new types are to be used - no sense in making a large bulk of changes if I have to retroactively make changes to the PR so that everyone feels comfortable in how it's used, and is most end-user-capable. :smile: 

As of now, these are stubbed in and aren't implemented in any of the type creation / constructors so that they can be used. I'd like to get the types narrowed down before getting the cons, etc., modified so they can be properly consumed.

@dsyme - @7sharp9 pointed me to this issue. If there's a more proper issue# I should be referencing, please let me know. At the moment they're simple stubbed types. If desired, these can be refactored to APs for better / more proper use, as you'd described. As he's best considered as a consumer of `FSharp.Compiler.Services`, and can give good feedback on how he'd be using it, I consider him my point-of-referral on usability. I will consider your suggestion for making these subtypes be handled by APs.

7sharp9:
I strongly disagree.  The symbol API was introduced to expose more details from the compiler about the context of a particular node in the code rather than a simple textual tip.

@dsyme RE: *(e.g. should "Active Pattern" be a subtype? I'd rather not go there)*
Active patterns and unions **are** already subtypes, see [here](https://github.com/fsharp/FSharp.Compiler.Service/blob/master/src/fsharp/vs/Symbols.fs#L1889)

Regarding `FSharpMemberOrFunctionOrValue` splitting theres an awful lot of internal pattern matching around the initial components `Event`, `Property`, `Member`, `Value`, and lots of mutually exclusive properties that could and do throw exceptions, using the API is a minefield.  Essentially the Symbol API publicly exposes what Item represents: https://github.com/fsharp/FSharp.Compiler.Service/blob/9a3fc667180ab0a04ee68d341804d40b9d34a008/src/fsharp/nameres.fs#L142-L194 coalescing some of these type into `FSharpMemberOrFunctionOrValue` and then having to re-expand them with active patterns while avoiding certain properties seems overly obtuse, especially when this API is supposed to expose an API that makes it easier for editors and tooling.

The actual issue for this would be
https://github.com/fsharp/FSharp.Compiler.Service/issues/363

dsyme:
> using the API is a minefield

I'm not disagreeing with that :)  However I'd rather we took the F# approach to clearing the minefield - for example, make all failure cases explicit as either "TryXYZ" methods/properties + option return values.  Creating a hierarchy in the Java/C# style brings plenty of its own problems.

FWIW the "active pattern cases" refer to the symbols for cases themselves, rather than the active pattern function itself.  The symbols for active pattern cases are very different to other symbols.

7sharp9:
Ok, I guess Im overruled :-(

Whats needed then is to hoist the AP's from XSF# and plug them in here with tests...

dsyme:
Closing this for now

jeroldhaas:
That's fine. I'll re-approach this at a later date.

PR #387, opened 2015-07-28
types and arrays in F# attribute contructor arguments

Jand42:
Enable reading attribute constructor argument values defined with `typeof` or array syntax. Value is a boxed `FSharpType` or boxed `obj[]`.

Notes:
I also experimented with correct handling `typedefof` but that would require some additional functions to look up `System.Tuple` and `FSharpFunc` cases properly.
Also, reading arrays and types from IL seemed to have some deeper issues in deserialization.

### DIFF
--- a/src/fsharp/infos.fs
+++ b/src/fsharp/infos.fs
@@ -2541,7 +2541,7 @@ let rec evalILAttribElem e =
     | ILAttribElem.TypeRef (Some _t) -> fail()
     | ILAttribElem.TypeRef None     -> null
 
-let evalFSharpAttribArg e = 
+let rec evalFSharpAttribArg g e = 
     match e with
     | Expr.Const(c,_,_) -> 
         match c with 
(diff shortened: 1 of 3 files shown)


src/fsharp/infos.fs
```diff
@@ -2560,8 +2560,9 @@ let evalFSharpAttribArg e =
         | Const.Zero -> null
         | Const.String s ->  box s
         | _ -> fail()
-    // TODO: typeof<..> in attribute values
-    // TODO: arrays in attribute values
+    | Expr.Op (TOp.Array,_,a,_) -> box [| for i in a -> evalFSharpAttribArg g i |]
+    | TypeOfExpr g ty -> box ty
```
dsyme: This use of "box" is unusual.  I suppose it does make for a minimal diff, which is good, but could you add a comment to evalFSharpAttribArg to indicate the range of objects that can be returned from the function?

dsyme:
One comment on the code - otherwise this looks great

PR #393, opened 2015-08-10
Update FAKE to 4.1 and all build dependencies

forki:
This one uses the latest FCS to allow F# 4.0 in build scripts. Also dogfood


dungpa:
While you're at it, could you help renovate the Release target?

Right now, NuGet push is not automatic; git tags and release notes may be out of sync with actual releases.

forki:
Yep will do

dsyme:
@forki - do you want to do that in a separate commit or as part of this one?

forki:
I think I do a second one

PR #394, opened 2015-08-10
Update build to latest and greates

forki:
as requested via https://github.com/fsharp/FSharp.Compiler.Service/pull/393#issuecomment-129428997

- [x] - Restore packages via paket
- [x] - tag git repo
- [x] - Create Github release via octokit
- [x] - Make AppVeyor run NuGet but not release target

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,4 @@ FSharp.Compiler.Tools.Nuget/*.nupkg
 FSharp.Core.Nuget/*.nupkg
 *.orig
 .paket/paket.exe
+paket-files
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,7 +2,7 @@ os: Windows Server 2012 R2
 init:
   - git config --global core.autocrlf input
 build_script:
-  - cmd: build.cmd Release
+  - cmd: build.cmd NuGet
 test: off
 version: '{build}'
 artifacts:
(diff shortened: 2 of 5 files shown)


forki:
@dungpa what do you mean with "NuGet push is not automatic"? 

https://github.com/fsharp/FSharp.Compiler.Service/blob/master/build.fsx#L152 shows that a publisher only needs to provide it's key. this is like the recommended projectscaffold way (see http://fsprojects.github.io/ProjectScaffold/release-process.html#Setup-release-script)

forki:
what else do you need?

dungpa:
@forki I often use `Publish = true`. I don't if it makes any difference.

Could we make `Release` depend on `ReleaseDocs`? At the moment, `ReleaseDocs` is often out of sync.

Could we use prompt users and passwords as per https://github.com/fsprojects/ProjectScaffold/pull/200?

dungpa:
@rneatherway Is there any problem with doing release using Octokit on Mono? I recalled you have to keep a private copy of Octokit for your projects.

rneatherway:
@dungpa the version of Octokit in FAKE master has an `isMono` switch to avoid the connection retry logic that crashes on Mono 4.0.2. So everything should be fine. The crash is fixed in Mono master so after the next release we should be able to reinstate the retry logic.

PR #405, opened 2015-08-14
Alignment and make internal lexer/parser properly internal

dsyme:
This makes the internal Lexer and Parser modules properly internal

This imposes a small source code incompatibility: after this change you must replace

    Parser.tagOfToken (Parser.token.IDENT "") 

by 

    FSharpTokenTag.Identifier


7sharp9:
@dsyme I just tested 1.4.1 on Xamarin studio and we now have breaking changes:

LexHelp is now internal, we use:
Lexhelp.Keywords.QuoteIdentifierIfNeeded
Lexhelp.Keywords.keywordNames

Parser is now internal we use this in the semantic highlighting for speed:

We use the function `Parser.tokenTagToTokenId`
As well as the `tokenId`'s
```
let token = Parser.tokenTagToTokenId ts.TokenInfo.Tag
        match token with
        | Parser.tokenId.TOKEN_LPAREN
        | Parser.tokenId.TOKEN_RPAREN
        | Parser.tokenId.TOKEN_LBRACK
        | Parser.tokenId.TOKEN_RBRACK
        | Parser.tokenId.TOKEN_LBRACE
        | Parser.tokenId.TOKEN_RBRACE 
        | Parser.tokenId.TOKEN_LBRACK_LESS
        | Parser.tokenId.TOKEN_GREATER_RBRACK
        | Parser.tokenId.TOKEN_LESS
        | Parser.tokenId.TOKEN_GREATER
        | Parser.tokenId.TOKEN_LBRACK_BAR
        | Parser.tokenId.TOKEN_BAR_RBRACK -> Some PunctuationBrackets
```

As long as this is internal we cannot upgrade.

dsyme:
Thanks. We can fix these by publishing alternative APIs that reveal just these. ( Feel free to send a PR for that if you need to expedite it) Are these all you rely on?  Tgx

7sharp9:
I think so yes, the tokentagtotokenid allows for a fast int search from the tokens rather than string based comparisons.  I think vfspt uses strings so it didn't affect them.

dsyme:
See https://github.com/fsharp/FSharp.Compiler.Service/pull/412 for a fix that makes more things public.  A set of token tag integers are made available so you don't need to use tokenTagToTokenId.

7sharp9:
Ok cool thanks, I'm having a bit of a holiday.  I'll check it out when I return 😀

dsyme:
@7sharp9 Enjoy your break!

PR #415, opened 2015-09-06
make Internal.Utilities.Text.Lexing.Position a struct

vasily-kirichenko:
Heap allocation before:
position-is-record
After:
position-is-struct-2

### DIFF
--- a/src/fsharp/LexFilter.fs
+++ b/src/fsharp/LexFilter.fs
@@ -403,13 +403,19 @@ type LexbufState(startPos: Position,
     member x.EndPos = endPos
     member x.PastEOF = pastEOF
 
+[<Struct>]
+type PositionTuple =
+    val X: Position
+    val Y: Position
+    new (x: Position, y: Position) = { X = x; Y = y }
+
 /// Used to save the state related to a token
 [<Class>]
 type TokenTup = 
     val Token : token
     val LexbufState : LexbufState
-    val LastTokenPos: Position * Position
-    new (token,state,lastTokenPos) = { Token=token; LexbufState=state;LastTokenPos=lastTokenPos }
+    val LastTokenPos: PositionTuple
+    new (token,state,lastTokenPos) = { Token=token; LexbufState=state; LastTokenPos=lastTokenPos }
     
     /// Returns starting position of the token
     member x.StartPos = x.LexbufState.StartPos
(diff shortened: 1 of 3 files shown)


dsyme:
Hi @vasily-kirichenko 

This is potentially a good change. A couple of things

* Although allocations will reduce, it's possible that the lexer and parser become slower (because more time is spent copying structs). Could you write a perf test to repeatedly parse some input?  I don't think it matters too much which input it is.

* Ideally this change would via  Microsoft/visualfsharp --> fsharp/fsharp --> fsharp/FSharp.Compiler.Service to keep the codebases maximally aligned. Would you be able to submit the change to Microsoft/visualfsharp and assess its impact on perf of a compiler bootstrap?

Thanks
Don

vasily-kirichenko:
Hi, Don.

How exactly I can run parsing on a file? If I use `FSharpChecker.ParseFileInProject`, it caches results and all subsequent calls returns immediately.

dsyme:
I guess you could generate a sequence of textually different files?

vasily-kirichenko:
I've tested it on all *.fs files in src/fsharp dir https://github.com/vasily-kirichenko/FSharp.Compiler.Service/blob/parser-perf/tests/ConsoleApplication1/Program.fs 

Results: before - 6.7 s, after - 7.03 s, which is roughly 5% digression :(   

Memory traffic (all objects created during the parsing), before:

before_memory

after:

after_memory

Now `Tuple<Position, Position>` takes almost 2x more memory. Maybe it's a good idea to replace it with a custom struct?

vasily-kirichenko:
I replaced a couple of top allocated `Tuple`s with custom structs and make a couple of functions to be top-level. After that parsing time reduced and equals 6.26 s (6% better than "before" result). Number of allocated objects reduced by 45% (!), allocated memory - by 9%. I tried to make `TokenTup` to be a struct, but it increased parsing time by 10%. 

after_2_memory

dungpa:
@dsyme Have you had a second look at this PR? It seems like a good enhancement to make.

dsyme:
> I replaced a couple of top allocated  Tuple s with custom structs and make a couple of functions to be top-level. After that parsing time reduced and equals 6.26 s (6% better than "before" result). Number of allocated objects reduced by 45% (!), allocated memory - by 9%. 

@vasily-kirichenko Is this what's currently in the PR?  Thx.

dsyme:
(I'll reopen since there does seem to be value here, once we confirm the tradeoffs)

vasily-kirichenko:
@dsyme yes.

dsyme:
@vasily-kirichenko Would it be possible for you to submit this to Microsoft\visualfsharp as well please?  I don't mind accepting it here first but ideally we should get it submitted there too.

vasily-kirichenko:
@dsyme done https://github.com/Microsoft/visualfsharp/pull/664

PR #418, opened 2015-09-08
Exposing assembly attributes on FSharpAssemblySignature

Jand42:
When checking a whole project, it would be useful to be able to get the list of attributes defined for the assembly. This is now exposed in the `projectResults.AssemblySignature.Attributes` property.

These top level attributes are not persisted during incremental type checking, and also with FSI they are appended to the dynamic assembly but not tracked. I did not change this, so the attribute list for a partial assembly signature is always empty.


dsyme:
Great work, thanks.

PR #428, opened 2015-10-04
Tool version numbers have changed in mono 4.2

ryansroberts:
Ugly, but should hopefully be backwards compatible

Related issue:
https://github.com/fsprojects/FSharpLint/issues/123

Also causes problems for fsharpautocomplete


dsyme:
Is there some test we could have added for this?

rneatherway:
@ryansroberts thanks. So mono has decided to decouple from the version numbers used by MS? Did you open an issue about this elsewhere? I seem to remember someone mentioning a problem in fsautocomplete like this.

PR #430, opened 2015-10-07
Exposing implemented abstract slots

Jand42:
New classes are added to expose SlotSig and SlotParam typed in Symbols API. Previously there was no way to tell which abstract slot an override or an interface implementation member were using.

Also there is a small fix included to have all assembly-level attributes visible on FSharpAssemblySignature.Attributes, not just the ones declared in the last code file.

### DIFF
--- a/src/fsharp/infos.fs
+++ b/src/fsharp/infos.fs
@@ -375,6 +375,9 @@ type ValRef with
          | TSlotSig(_,oty,_,_,_,_) :: _ -> isInterfaceTy g oty
          | [] -> false)
 
+    member vref.ImplementedSlotSignatures =
+        vref.MemberInfo.Value.ImplementedSlotSigs
+
 //-------------------------------------------------------------------------
 // Helper methods associated with using TAST metadata (F# members, values etc.) 
 // as backing data for MethInfo, PropInfo etc.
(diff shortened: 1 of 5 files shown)


7sharp9:
The terminology in the symbols API is strange, "abstract slot signatures" sounds far more technical than it needs to.

Jand42:
@7sharp9 Can you suggest some better naming? Thanks!

src/fsharp/infos.fs
```diff
@@ -375,6 +375,9 @@ type ValRef with
          | TSlotSig(_,oty,_,_,_,_) :: _ -> isInterfaceTy g oty
          | [] -> false)
 
+    member vref.ImplementedSlotSignatures =
+        vref.MemberInfo.Value.ImplementedSlotSigs
```
dsyme: The "Value" might fail here (MemberInfo is an option).  When the value is not a member then "ImplementedSlotSigs" in the symbols API should I think return an empty list

src/fsharp/vs/Symbols.fs
```diff
@@ -855,6 +855,57 @@ and FSharpDelegateSignature(cenv, info : SlotSig) =
         | Some ty -> FSharpType(cenv,  ty)
     override x.ToString() = "<delegate signature>"
 
+and FSharpAbstractParameter(cenv, info : SlotParam) =
```
dsyme: It's a slight shame to require a new API type here (rather than FSharpParameter) - but I suppose it's the natural way to do it and sufficient.

dsyme:
This looks good.  I'm surprised at the AppVeyor build failure w.r.t. SourceLink  - ideally that part of the build shouldn't be being run on appveyor.

dsyme:
@Jand42  Could you please add tests for the abstract slot signatures? Thanks

Jand42:
@dsyme Added some tests and fixed possibly failing `.Value`. Thanks for the review.

Jand42:
Sorry for the extra commits.

dsyme:
Thanks for the great contribution

PR #431, opened 2015-10-09
Fix symbol types where an incomplete type check occurs

7sharp9:
The PR fixed the fragmented types that are currently presented on an
incomplete type check i.e. if a referenced nuget package is not
restored etc.

e.g. types may be represented by `?25354-> ?25345`  rather than ’a -> ‘b

Ive fixed all the occurrence that I can see happen although there could
be further issues in `FSharpMemberOrFunctionOrValue.FullType` as I only
touched `ValRef`’s

### DIFF
--- a/src/fsharp/vs/Symbols.fs
+++ b/src/fsharp/vs/Symbols.fs
@@ -1100,7 +1100,8 @@ and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) =
                 let rty = m.GetFSharpReturnTy(cenv.amap,range0,m.FormalMethodInst)
                 let argtysl = m.GetParamTypes(cenv.amap,range0,m.FormalMethodInst) 
                 mkIteratedFunTy (List.map (mkTupledTy cenv.g) argtysl) rty
-            | V v -> v.TauType
+            | V v -> let _, typ, _ = PrettyTypes.PrettifyTypes1 cenv.g v.TauType
+                     typ
         FSharpType(cenv,  ty)
 
     member __.HasGetterMethod =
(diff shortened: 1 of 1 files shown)


7sharp9:
Any comments?

src/fsharp/vs/Symbols.fs
```diff
@@ -1460,8 +1470,8 @@ and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) =
             let tau = v.TauType
             let argtysl,_ = GetTopTauTypeInFSharpForm cenv.g curriedArgInfos tau range0
             let argtysl = if v.IsInstanceMember then argtysl.Tail else argtysl
-            
-            [ for argtys in argtysl do 
+            let prettyArgtysl = argtysl |> List.map (fun a -> PrettyTypes.PrettifyTypesN1 cenv.g (a, tau))
```
dsyme: You should prettify tau, not the pieces one by one.

src/fsharp/vs/Symbols.fs
```diff
@@ -1430,8 +1432,15 @@ and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) =
...
             
             [ for argtys in m.GetParamDatas(cenv.amap,range0,m.FormalMethodInst) do 
-                 yield 
-                   [ for (ParamData(isParamArrayArg,isOutArg,optArgInfo,nmOpt,_reflArgInfo,pty)) in argtys do 
+                 let _, prettyTyps, _cxs =
+                     argtys
+                     |> List.map (function ParamData(isParamArrayArg,isOutArg,optArgInfo,nmOpt,_reflArgInfo,pty) -> pty)
+                     |> PrettyTypes.PrettifyTypesN cenv.g
+                 let combined =
+                   List.map2 (fun pty (ParamData(isParamArrayArg,isOutArg,optArgInfo,nmOpt,reflArgInfo,_pty)) ->
```
dsyme: Please use ``(prettyTypes, argtys) ||> List.map2 (fun ... -> ...)``

src/fsharp/vs/Symbols.fs
```diff
@@ -1100,7 +1100,8 @@ and FSharpMemberOrFunctionOrValue(cenv, d:FSharpMemberOrValData, item) =
                 let rty = m.GetFSharpReturnTy(cenv.amap,range0,m.FormalMethodInst)
                 let argtysl = m.GetParamTypes(cenv.amap,range0,m.FormalMethodInst) 
                 mkIteratedFunTy (List.map (mkTupledTy cenv.g) argtysl) rty
-            | V v -> v.TauType
+            | V v -> let _, typ, _ = PrettyTypes.PrettifyTypes1 cenv.g v.TauType
```
dsyme: The need to prettify types can arise not only for functions and values but also for members, properties and events.  You should probably do those here too

dsyme:
It looks good.  Unit tests?

7sharp9:
Unit test will require a discrete fragmented type failure case, the only one I know of is when a nuget package is not restored, Ill try and get a small unit test to cover this.

dsyme:
@7sharp9 This kind of problem should occur whenever necessary referenced DLLs are not available and you have code that uses constructs in those DLLs.

src/fsharp/vs/Symbols.fs
7sharp9: Not sure I understand, `PrettyTypes.PrettifyTypes1 tau` ?
Would we not have to do `GetTopTauTypeInFSharpForm` and then disregard the head if its an instance?

dsyme:
Closed in favour of #442

PR #432, opened 2015-10-10
Integrate with fsharp/fsharp

rojepp:


dsyme:
Hi @rojepp - I think something has happened to the commit history....there should be more than one commit, something like this : https://github.com/fsharp/fsharp/pull/498/commits

rojepp:
@dsyme Seems I messed up. Closing this.

PR #433, opened 2015-10-10
Integrate with fsharp/fsharp

rojepp:
This time with preserved history.


dsyme:
This looks great.  Any specific concerns?

rojepp:
No, not really. A few false starts is all. :)
I'm fairly confident that it went well.

dsyme:
Like it, thanks for the great contribution in doing this integration!!

PR #435, opened 2015-10-11
F sharp merge 10 2015 b

brodyberg:
Gave this another shot by following @rojepp's suggestion of using Beyond Compare. Even with that I found it difficult to assess which of the three changes to take: HEAD, BASE or FSharp.


dsyme:
Something also went wrong with this - there are too many commits in the history here.  Will take a look at the command you use in the other PR

PR #443, opened 2015-10-15
Add ability to downsize caches in response to a max-memory signal

dsyme:
Make FCS automatically downsize caches when a maximum allocated memory threshold memory is reached.

This is somewhat experimental but may be the best solution to the "Visual Studio grinds to a complete halt" issue discussed here: https://github.com/fsprojects/VisualFSharpPowerTools/issues/1160#issuecomment-148352542

The default threshold is 1.7GB of allocated managed memory in a process.  This corresponds to a managed memory size of around 2.5GB.  When this happens, the "strong" values for keeping background project builds are reduced to size 1 for the remainder of the lifetime of the FSharpChecker object.  In practice this will still make Visual Studio F# Power Tools usable, but some operations like renaming across multiple projects may take longer.

Reducing the FCS strong cache sizes does _not_ guarantee there will be enough memory to continue operations - even holding one project strongly may exceed a process memory budget.  It just means FCS may hold less memory strongly.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,8 @@
+#### 1.4.0.8 - 
+* FSharpType.Format now prettifies type variables.  If necessary, FSharpType.Prettify can also be called
+* Add maximum-memory trigger to downsize FCS caches. Defaults to 1.7GB of allocaed memory in the system 
+  process for a 32-bit process, and 2x this for a 64-bit process
+
 #### 1.4.0.7 - 
 * fix 427 - Make event information available for properties which represent first-class uses of F#-declared events
 * fix 410 - Symbols for C# fields (and especially enum fields)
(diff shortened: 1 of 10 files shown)


vasily-kirichenko:
Looks good! 

I'm worried about caches size = 1. I experimented with `projectCacheSize = 5` on real solution and for some reason no VFPT features worked. It's strange, maybe I was wrong somewhere. I'll test this PR.

dsyme:
@vasily-kirichenko Right now I can't even edit at all when this kind of memory usage is reached - all of Visual Studio grinds to a halt.  So even if VFPT features stopped working this would actually still be an improvement, though far from ideal

vasily-kirichenko:
@dsyme I'm afraid you have to publish new FCS NuGet, otherwise CI cannot build VFPT.

dsyme:
https://www.nuget.org/packages/FSharp.Compiler.Service/1.4.0.8

vasily-kirichenko:
https://github.com/fsprojects/VisualFSharpPowerTools/pull/1176

vasily-kirichenko:
CI seems to slow at the moment, I've shared the VSIX here https://docs.google.com/uc?authuser=0&id=0B8HLQUKik9VtS0RTN1F3NzFJbEE&export=download

PR #445, opened 2015-10-16
Fix caching, rewrite FCS reactor queue

dsyme:
Title says it. The caches were being missed in many cases, and the lookup of the caches was often asynchronously disconnected from the code to update that cache.

Also add tracing to get intelligible info about the request queue via Debug.WriteLine.  This goes to the output window of Visual Studio. For example when tracing the request queue for VFPT you can start an instance of devenv.exe under another VS debugger:

     devenv /debugexe devenv.exe

However this currently requires using a debug build of FSharp.Compiler.Service

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,7 @@
 * Write trace information about the reactor queue to the event log
 * Rewrite reactor to consistently prioritize queued work
 * Implement cancellation for queued work if it is cancelled prior to being executed
+* Adjust caching to check cache correctly if there is a gap before the request is executed
 
 #### 1.4.0.9 - 
 * FSharpType.Format fix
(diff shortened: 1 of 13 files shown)


7sharp9:
Would this potentially help with the dirty notify issues Ive mentioned before?  I have no idea of the context which might of changed when I do get the notifications they are just normal parse requests.

dsyme:
I know you sent some details by email, but could you enter that as an issue in this repo, with repro steps or a failing test?

vasily-kirichenko:
I tried this build on VFPT solution and VS's memory reached 2.3GB in 20 minutes :(

dsyme:
@vasily-kirichenko I wasn't expecting this one to fix any specific memory problems.

PR #447, opened 2015-10-16
fix bug with double lookup of cache

dsyme:
There was a bug in the AgedLookup code - if you lookup the cache twice it replaces the key in the entry on the first lookup, giving incorrect results on the second lookup


vasily-kirichenko:
It's really great. Can you publish a new nuget, please?

PR #453, opened 2015-10-22
Repo for bug that affects FAKE

robertpi:
Both scripts should parse & execute correctly, the second one doesn't.

Suspect it might relate to an optimization to use constants rather than a let binding for strings, but can't find such a change. Maybe a bisect would help, but not had chance to try that.


dsyme:
A fix for this is here: https://github.com/Microsoft/visualfsharp/pull/756   We should integrate this into this repository

dsyme:
The fix has been integrated

PR #457, opened 2015-11-04
Change throw/catch to conditional. fixes #456

### LINKED ISSUE #456: FCS makes calls to kernel32.dll when running on OSX
https://github.com/fsharp/FSharp.Compiler.Service/blob/c1ca06144d8194000cf6b86f5f26bdc433ccaa7d/src/absil/ilread.fs#L169

screen shot 2015-11-03 at 15 08 51

nosami:


src/absil/ilread.fs
```diff
@@ -3971,10 +3971,7 @@ let ClosePdbReader pdb =
 #endif
 
 let OpenILModuleReader infile opts = 
-
-#if FX_NO_NATIVE_MEMORY_MAPPED_FILES
-#else
-   try 
+    if not IL.runningOnMono then
```
dsyme: runningOnMono isn't what we want here, because we may be running with Mono on Windows.

I think just check for Windows.

nosami:
Not sure about the Windows check. I assume it's right - I copied it from https://github.com/fsharp/FSharp.Compiler.Service/blob/conditional-memory-mapped-files/samples/FscExe/FscMain.fs#L45

dsyme:
Great thanks

Issue #371, opened 2015-07-09
Discussion for implementing checker.GetProjectOptionsFromProjectFile in a separate process


7sharp9:
Issues #366 and #342 both require either assembly redirects in a seperate process or some form of reflective binding.  I created this issue to discus how we want this implementing.

* How do we want to proceed with this reflection or a separate process being called with assembly redirects
*  Do we want to add an extra parameter to the existing function to force running in a separate process, add a new function, or simply update the current function to always run in a separate process
* Running via a separate process seems the simplest option but how to we represent the output data

Rather than a separate process, maybe we could use `AppDomainSetup.ConfigurationFile` and `AppDomain.CreateDomain()` before running `checker.GetProjectOptionsFromProjectFile` in the new appdomain.

//cc @dsyme @rneatherway

rneatherway:
> How do we want to proceed with this reflection or a separate process being called with assembly redirects

How do you set assembly redirects on a new process?

> Do we want to add an extra parameter to the existing function to force running in a separate process, add a new function, or simply update the current function to always run in a separate process

I think it is fine to always run in a new process.

> Running via a separate process seems the simplest option but how to we represent the output data

You said before that XS runs the resolution in a separate process, would you mind linking to that (sorry if you already did before)? I'm wondering how they handle sending the data back and forth.

> Rather than a separate process, maybe we could use AppDomainSetup.ConfigurationFile and AppDomain.CreateDomain() before running checker.GetProjectOptionsFromProjectFile in the new appdomain.

Nice idea, but I don't really know anything about AppDomains. Do they provide sufficient isolation?

7sharp9:
I believe the format is normal remoting.

It uses a `RemoteBuildEngine`: https://github.com/mono/monodevelop/blob/5436cd09a80af3eb9da1200c76a9c410c03bf7d3/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.Formats.MSBuild/RemoteProjectBuilder.cs#L39

Which is created here:
https://github.com/mono/monodevelop/blob/6e87c2c761c7bf22b2c079f9818c7f4967b3bb1f/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.Formats.MSBuild/MSBuildProjectService.cs#L504

In this instance the process thats invoked (MonoDevelop.Projects.Formats.MSBuild.exe) has an assembly redirect in its config.

rneatherway:
[...]

Issue #354, opened 2015-06-19
`GetOptionsFromProjectFile` returns duplicates of project references on mono


rneatherway:
This is essentially benign as they seem to be ignored, but should still be fixed.

dsyme:
ugh, thanks for detecting all these!!!

rneatherway:
No longer the case following #478

PR #459, opened 2015-11-06
Perform project cracking out of process

rneatherway:
References: #371, #354, #342, #337.

I've exposed the logging a bit more easily as when using this it is really important to be able to get feedback from users as to why it has broken again on their system.

This causes the project options tests for type providers to fail, but everything else is working. I'd appreciate some help if there is any magic I'm missing here, like if the new assembly needs a property set or something. (/cc @dsyme)

Example:

```
***** FSharp.Compiler.Service.Tests.ProjectAnalysisTests.Test Project25 whole project errors
[|startup (1,1)-(1,81) parameter warning Referenced assembly '/home/robin/dev/other/FSharp.Compiler.Service/tests/service/FSharp.Data.dll' has assembly level attribute 'Microsoft.FSharp.Core.CompilerServices.TypeProviderAssemblyAttribute' but no public type provider classes were found;
  /tmp/tmp56ff41d1.fs (4,16)-(4,27) typecheck error The type 'XmlProvider' is not defined;
  /tmp/tmp56ff41d1.fs (4,16)-(4,27) typecheck error The type 'XmlProvider' is not defined;
  /tmp/tmp56ff41d1.fs (5,17)-(5,26) typecheck error The field, constructor or member 'GetSample' is not defined;
  /tmp/tmp56ff41d1.fs (10,9)-(10,20) typecheck error The value or constructor 'XmlProvider' is not defined|]
```

Once that is fixed, I will do some tidying as below:

Todo:
- [x] Remove ProjectCrackerReader, used for testing prior to hooking up to FCS
- [x] Remove FCS references of MSBuild v12 and FSharpProjectFileInfo from service.fs
- [x] Change tests to use GetProjectOptionsFromProjectFile


7sharp9:
Are you adding the `mono` prefix when starting the process?

rneatherway:
No, mono's StartProcess does that automatically if the target is a CLR executable.

dsyme:
I couldn't immediately spot why that type provider DLL would be causing problems given your changes - nothing's really changed unless the actually options cracked have changed (e.g. the FSHarp.Core being chosen has changed - that could cause the error you're seeing).  Are tests running clean without your changes?  thanks

rneatherway:
@dsyme yes, the tests run clean without the changes. I'll have to get logs for those projects before and after for comparison.

rneatherway:
@dsyme I'd like to ask a quick bit of advice. I have this all working quite nicely, but FSharp.Core is proving to be a problem. This new `ProjectCracker.exe` needs to be in the same directory as `FSharp.Compiler.Service.dll` so that `FSharp.Compiler.Service.dll` can access the record type that is used to serialize the results. However that means there can't be an FSharp.Core in that directory (because it is the choice of the consumer of FCS what FSharp.Core they should bind against). I then can't see how to make this binary work on machines that may have only one of FSharp.Core 4.3.0.0, 4.3.1.0 or 4.4.0.0. As it is essentially just a single-file application, should I convert it to C#?

dsyme:
I see...  Could you put it in a sub-directory and resolve the FCS assembly reference to the shared component via an AssemblyLoad event resolver?

dsyme:
Mind you it's a bit ugly to have FCS need a subdirectory.  Ugh.

dsyme:
You could statically link FSharp.Core I guess.  It seems really unfortunate to rewrite it in C#, I'd prefer not to have C# in such a very base level F# tooling component. 

BTW could you call it FSharp.Compiler.Service.ProjectCracker.exe?

dsyme:
BTW it's beginning to feel like the project cracking functionality should go in a new different nuget package,  FSharp.Compiler.Service.ProjectCracker, with a DLL and the .EXE tool (FSharp.Compiler.Service.ProjectCracker.Tool.exe or something) .  Then projects that need the functionality can opt-in to it and won't be surprised to see these extra files.

rneatherway:
At the moment it is called `FSharp.Compiler.Service.ProjectCracker.exe`, I just called it `ProjectCracker.exe` for short. Currently the dependency is that `FCS.dll` references `FCSPC.exe`, but that could be reversed, or a tiny shared assembly could provide the common types for the two to talk to each other.

> I see... Could you put it in a sub-directory and resolve the FCS assembly reference to the shared component via an AssemblyLoad event resolver?

Yes, I thought about this, but I wasn't sure exactly how it would work. I suppose the resolver could scan Assembly.Load from a file on disk. I also thought it would be really ugly though.

> You could statically link FSharp.Core I guess.

Right, I guess I could depend on the FSharp.Core NuGet package and get it that way, then statically link. Worth a shot.

> BTW it's beginning to feel like the project cracking functionality should go in a new different nuget package, FSharp.Compiler.Service.ProjectCracker, with a DLL and the .EXE tool (FSharp.Compiler.Service.ProjectCracker.Tool.exe or something) .

What DLLs exactly are you thinking should go in here? Should it be completely separate from FCS? The two need to talk to each other somehow to get the correct types (I want it to return `FSharpProjectOptions`).

dsyme:
@rneatherway I'm thinking that the API functionality to crack project options be removed from FSharp.Compiler.Service.dll and moved into a new DLL (in a new package), which depends on FCS.dll, and that new package also contain ProjectCracker.exe.  FCS.dll would not depend on either of these. This would be an API change but we'd mark the FCS APIs as deprecated.  

The new package would still be in this repo and part of the default build & test.

dsyme:
In that scenario I mind less if ProjectCracker.exe is implemented in C# (since it leaves the primary FCS.dll entirely in F# with a low dependency base)

rneatherway:
I appreciate the goal of a clean separation between FCS and the project cracking, but I'm not sure this idea actually achieves that. If we leave the old API in place, albeit deprecated, then the FCS dependency on MSBuild v12 remains, which causes these installation problems in some of the referenced issues. My current approach is the only way that prevents API breakage *and* removes the dependency on MSBuild v12 for FCS.dll.

I'd like to propose the following:

* Use JSON serialization to break the binary dependency between the two projects. Presumably this means using something built into .NET as we don't want FCS depending on other NuGet packages.
* Static-link FSharp.Core into FCSPR.exe, or FCSPR.exe to C# (No C# code would end up in FCS.dll). Then FCSPR.exe can remain in the same folder as FCS.dll.
* Leave FCSPR.exe in the same NuGet package. This is necessary to prevent API breakage.

What do you think?

7sharp9:
We've never had any problems calling the project cracking as an executable, no FSharp.Core issues.

rneatherway:
Which FSharp.Core do you bundle with the addin?

7sharp9:
Hmmm the latest I think, so 4.4.0.0

rneatherway:
Right, I think that's part of the issue. FCS isn't allowed to bundle FSharp.Core, hence the difficulty just being able to accept any version that might be lying around.

PR #463, opened 2015-11-12
Exposing slot signature of members in object expressions

Jand42:
A small addition of a `.Signature` member to `FSharpObjectExprOverride`. This is needed for identifying the member overridden or implemented.

There are no current tests included for `FSharpObjectExprOverride` as I've seen, but I tested this change in a bigger project and it worked correctly.

### DIFF
--- a/src/fsharp/vs/Exprs.fs
+++ b/src/fsharp/vs/Exprs.fs
@@ -108,7 +108,8 @@ type E =
     | ILAsm of string * FSharpType list * FSharpExpr list
 
 /// Used to represent the information at an object expression member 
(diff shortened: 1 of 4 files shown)


dsyme:
Hmmm looks like Travis CI is busted - perhaps something to do with using Mono 4.0.5.1

dsyme:
The change looks good - any chance you could add some tests too? I know you said there weren't any for  FSharpObjectExprOverride  but please add at least a smoke test for that, along with this new addition. thanks

Jand42:
@dsyme Thanks, I will do this, probably tomorrow.

dsyme:
Thank you for adding the tests

PR #464, opened 2015-11-17
Move FCS project cracking out of process

rneatherway:
This aims to fix many problems stemming from two main issues:

* FCS depends on MSBuild v12, and not all machines have this installed.
* It's hard to apply robust assembly redirects, as this is up to the
  application referencing FCS. These are needed to redirect MSBuild v4
  to whatever is being used (for custom MSBuild tasks), and also to
  allow redirecting to MSBuild v14 if necessary.

The API is preserved, with an additional function,
`GetProjectOptionsFromProjectFileLogged` to allow easier access to the
logs for presentation to users and bug-reporting.

The functionality is implemented by FCS calling a new console
application, FSharp.Compiler.Service.ProjectCracker.exe, and the
response is transmitted via serialization using JSON to types shared by
a common source file, so there is no binary dependency between the two.

The ProjectCracker static links to FSharp.Core, as it is in the same
directory as FCS.dll. Thus bundling a separate copy would force the
choice of version on all users of FCS, which is not acceptable.

@dsyme this implements my suggestion from #459.

### DIFF
--- a/src/fsharp/FSharp.Compiler.Service.ProjectCracker/FSharp.Compiler.Service.ProjectCracker.fsproj
+++ b/src/fsharp/FSharp.Compiler.Service.ProjectCracker/FSharp.Compiler.Service.ProjectCracker.fsproj
@@ -1,0 +1,95 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <SchemaVersion>2.0</SchemaVersion>
+    <ProjectGuid>b1bdd96d-47e1-4e65-8107-fbae23a06db4</ProjectGuid>
+    <OutputType>Exe</OutputType>
+    <RootNamespace>FSharp.Compiler.Service.ProjectCracker</RootNamespace>
+    <AssemblyName>FSharp.Compiler.Service.ProjectCracker</AssemblyName>
+    <TargetFrameworkVersion Condition=" '$(TargetFrameworkVersion)' == '' ">v4.5</TargetFrameworkVersion>
+    <TargetFSharpCoreVersion>4.3.0.0</TargetFSharpCoreVersion>
+    <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\..\..\</SolutionDir>
+    <Name>FSharp.Compiler.Service.ProjectCracker</Name>
+    <OtherFlags>$(OtherFlags) --staticlink:FSharp.Core</OtherFlags>
+    <NoWarn>$(NoWarn);40</NoWarn>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <Tailcalls>false</Tailcalls>
+    <OutputPath>bin\Debug\</OutputPath>
+    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <WarningLevel>3</WarningLevel>
+    <PlatformTarget>AnyCPU</PlatformTarget>
+    <DocumentationFile>bin\Debug\FSharp.Compiler.Service.ProjectCracker.XML</DocumentationFile>
+    <Prefer32Bit>true</Prefer32Bit>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <Tailcalls>true</Tailcalls>
+    <OutputPath>bin\Release\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <WarningLevel>3</WarningLevel>
+    <PlatformTarget>AnyCPU</PlatformTarget>
+    <DocumentationFile>bin\Release\FSharp.Compiler.Service.ProjectCracker.XML</DocumentationFile>
+    <Prefer32Bit>true</Prefer32Bit>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="Microsoft.Build.Framework" Condition=" '$(TargetFrameworkVersion)' == 'v4.0'" />
+    <Reference Include="Microsoft.Build.Engine" Condition=" '$(TargetFrameworkVersion)' == 'v4.0'" />
+    <Reference Include="Microsoft.Build" Condition=" '$(TargetFrameworkVersion)' == 'v4.0'" />
+    <Reference Include="Microsoft.Build.Utilities.v4.0" Condition=" '$(TargetFrameworkVersion)' == 'v4.0'" />
+    <Reference Include="Microsoft.Build.Framework, Version=12.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" Condition=" '$(TargetFrameworkVersion)' == 'v4.5'">
+      <SpecificVersion>True</SpecificVersion>
+    </Reference>
+    <Reference Include="Microsoft.Build.Engine, Version=12.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" Condition=" '$(TargetFrameworkVersion)' == 'v4.5'">
+      <SpecificVersion>True</SpecificVersion>
+    </Reference>
+    <Reference Include="Microsoft.Build, Version=12.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" Condition=" '$(TargetFrameworkVersion)' == 'v4.5'">
+      <SpecificVersion>True</SpecificVersion>
+    </Reference>
+    <Reference Include="Microsoft.Build.Utilities.v12.0" Condition=" '$(TargetFrameworkVersion)' == 'v4.5'" />
+    <Reference Include="mscorlib" />
+    <Reference Include="FSharp.Core, Version=$(TargetFSharpCoreVersion), Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
+      <Private>False</Private>
+    </Reference>
+    <Reference Include="System" />
+    <Reference Include="System.Core" />
+    <Reference Include="System.Numerics" />
+    <Reference Include="System.Runtime.Serialization" />
+    <Reference Include="System.Xml" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="ProjectCrackerOptions.fs" />
+    <Compile Include="Program.fs" />
+    <None Include="App.config" />
+  </ItemGroup>
+  <PropertyGroup>
+    <MinimumVisualStudioVersion Condition="'$(MinimumVisualStudioVersion)' == ''">11</MinimumVisualStudioVersion>
+  </PropertyGroup>
+  <Choose>
+    <When Condition="'$(VisualStudioVersion)' == '11.0'">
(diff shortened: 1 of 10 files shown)


dsyme:
Great! :) Can you add a second nuget package called FSharp.Compiler.Service.ProjectCracker to the build?

src/fsharp/FSharp.Compiler.Service.ProjectCracker/FSharp.Compiler.Service.ProjectCracker.fsproj
```diff
@@ -0,0 +1,95 @@
...
+    <PlatformTarget>AnyCPU</PlatformTarget>
+    <DocumentationFile>bin\Release\FSharp.Compiler.Service.ProjectCracker.XML</DocumentationFile>
+    <Prefer32Bit>true</Prefer32Bit>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="Microsoft.Build.Framework" Condition=" '$(TargetFrameworkVersion)' == 'v4.0'" />
+    <Reference Include="Microsoft.Build.Engine" Condition=" '$(TargetFrameworkVersion)' == 'v4.0'" />
+    <Reference Include="Microsoft.Build" Condition=" '$(TargetFrameworkVersion)' == 'v4.0'" />
+    <Reference Include="Microsoft.Build.Utilities.v4.0" Condition=" '$(TargetFrameworkVersion)' == 'v4.0'" />
+    <Reference Include="Microsoft.Build.Framework, Version=12.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" Condition=" '$(TargetFrameworkVersion)' == 'v4.5'">
```
dsyme: What are the ramifications of having these version dependencies?

src/fsharp/FSharp.Compiler.Service.ProjectCracker/Program.fs
```diff
@@ -0,0 +1,414 @@
...
+
+    snd (getOptions file)
+
+  let addMSBuildv14BackupResolution () =
+    let onResolveEvent = new ResolveEventHandler(fun sender evArgs ->
+      let requestedAssembly = AssemblyName(evArgs.Name)
+      if requestedAssembly.Name.StartsWith("Microsoft.Build") &&
+          not (requestedAssembly.Name.EndsWith(".resources")) then
+        requestedAssembly.Version <- Version("14.0.0.0")
+        Assembly.Load (requestedAssembly)
```
dsyme: Can this go in a loop?  What happens if 14.0.0.0 is not available?

rneatherway:
Well, I'm not sure if I was quite clear [in the other thread](https://github.com/fsharp/FSharp.Compiler.Service/pull/459#issuecomment-155797118), but this replaces the API function `GetProjectOptionsFromProjectFile` in place. I think this is the best way forward, and in that case I'm not sure what would go in the other NuGet package.

src/fsharp/FSharp.Compiler.Service/FSharp.Compiler.Service.fsproj
```diff
@@ -588,23 +591,13 @@
...
-    </Reference>
-    <Reference Include="Microsoft.Build.Engine, Version=12.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" Condition=" '$(TargetFrameworkVersion)' == 'v4.5'">
-      <SpecificVersion>True</SpecificVersion>
-    </Reference>
-    <Reference Include="Microsoft.Build, Version=12.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" Condition=" '$(TargetFrameworkVersion)' == 'v4.5'">
-      <SpecificVersion>True</SpecificVersion>
-    </Reference>
-    <Reference Include="Microsoft.Build.Utilities.v12.0" Condition=" '$(TargetFrameworkVersion)' == 'v4.5'" />
-    <Reference Include="Microsoft.Build.Tasks.v12.0" Condition=" '$(TargetFrameworkVersion)' == 'v4.5'" />
+    <Reference Include="Microsoft.Build.Framework" />
```
dsyme: Why do these versions change for net45?  Won't this affect the behaviour of the ReferenceResolution.fs code in FCS?

tests/service/data/ToolsVersion12.fsproj
```diff
@@ -41,13 +41,6 @@
   <ItemGroup>
     <Compile Include="main.fs" />
   </ItemGroup>
-  <ItemGroup>
-    <ProjectReference Include="..\Test\test.fsproj">
-      <Name>Interfaces</Name>
-      <Project>{00000000-0000-0000-0000-000000000001}</Project>
-      <Private>True</Private>
-    </ProjectReference>
-  </ItemGroup>
```
dsyme: What's this change?

src/fsharp/vs/service.fs
```diff
@@ -3156,19 +2802,45 @@ type FSharpChecker(projectCacheSize, keepAssemblyContents, keepAllBackgroundReso
...
+        let p = new System.Diagnostics.Process()
+        p.StartInfo.FileName <- Path.Combine(Path.GetDirectoryName(Reflection.Assembly.GetExecutingAssembly().Location),
+                                             "FSharp.Compiler.Service.ProjectCracker.exe")
+        p.StartInfo.Arguments <- arguments.ToString()
+        p.StartInfo.UseShellExecute <- false
+        p.StartInfo.CreateNoWindow <- true
+        p.StartInfo.RedirectStandardOutput <- true
+        ignore <| p.Start()
+    
+        let ser = new System.Runtime.Serialization.Json.DataContractJsonSerializer(typeof<FSharp.Compiler.Service.ProjectCracker.ProjectOptions>)
```
dsyme: Might be good to have a test where a file name includes some Unicode characters.

src/fsharp/FSharp.Compiler.Service.ProjectCracker/Program.fs
```diff
@@ -0,0 +1,414 @@
...
+
+      let options = { ProjectFile = file
+                      Options = Array.ofList parsedProject.Options
+                      ReferencedProjectOptions = referencedProjectOptions
+                      LogOutput = parsedProject.LogOutput }
+      parsedProject.OutputFile, options
+
+    snd (getOptions file)
+
+  let addMSBuildv14BackupResolution () =
```
dsyme: Could you add a comment to this function explaining why dynamic resolution of binding is needed?  I understand why - it's just for posterity. thanks

src/fsharp/FSharp.Compiler.Service.ProjectCracker/Program.fs
rneatherway: No, each event handler (here only one) is called in turn, if they all return `null` then an exception is raised as normal as if there had been no event handlers in place.

src/fsharp/FSharp.Compiler.Service.ProjectCracker/FSharp.Compiler.Service.ProjectCracker.fsproj
rneatherway: I'm not 100% sure. I only envisage using the version 12 dependencies. I just copied the pattern used in FCS.fsproj for the past while. I'm testing without it now.

src/fsharp/FSharp.Compiler.Service/FSharp.Compiler.Service.fsproj
rneatherway: This was only changed when the project cracking capability was added, I'm just reverting it to how it was before.

tests/service/data/ToolsVersion12.fsproj
rneatherway: This file was invalid (the referenced project doesn't exist), so the test was failing for a reason unrelated to what it should test for.

src/fsharp/vs/service.fs
rneatherway: Great idea.

src/fsharp/FSharp.Compiler.Service.ProjectCracker/Program.fs
rneatherway: Will do.

dsyme:
> Well, I'm not sure if I was quite clear in the other thread, but this replaces the API function  GetProjectOptionsFromProjectFile  in place. I think this is the best way forward, and in that case I'm not sure what would go in the other NuGet package. 

I'm somewhat loathe to add a second binary (ProjectCracker.EXE) to the FSharp.Compiler.Service package. For some reason I would really like the FSharp.Compiler.Service package to remain a one-DLL nuget package. For example, we will eventually want to make FCS a .NET Standard or Portable component, and for those it may not make sense to have functionality that depends on using the .EXE. I guess we could just make the FCS API smaller on some targets, but I'd rather not.

I guess this is why it feels like the project cracking functionality inn the API should go in a separate package and DLL that has a .NET 4.x dependency - for now the same as MSBuild - and that package would also contain ProjectCracker.exe. 

Is there a particular reason not to create two packages like this (beyond it being work and involve a major version revision to FCS)? 

Thanks for your patience :)

rneatherway:
> Thanks for your patience :)

No problem, I'd rather work out the best way forward. As far as I can tell there are three options, but I'm not sure which you are proposing.

1. What I have submitted here. An in-place replacement for the existing FCS project cracking API. A new exe would be added to the NuGet package. I understand your objections to this option, and think they are valid.

2. No change to FCS, but a new NuGet package containing the new project cracking functionality. This doesn't solve the issue with MSBuild dependencies raised in the issues I referenced in the other PR, but also doesn't break the API or add new assemblies to the FCS NuGet package. My least favourite option.

3. Remove the project cracking from FCS.dll completely. Move it into a new dll/exe pair that live in a new NuGet package. This would break the FCS API, but a clean break. I'm also OK with this option. This package could also depend on FCS in order to return `FSharpProjectOptions` from FCS.

In the other PR I thought you were suggesting 2, but are you actually suggesting 3? And if so, should this new package depend on FCS? In this case I would probably also go back to the binary serialization.

src/fsharp/FSharp.Compiler.Service.ProjectCracker/Program.fs
```diff
@@ -0,0 +1,414 @@
...
+              do log.Verbosity <- Microsoft.Build.Framework.LoggerVerbosity.Diagnostic
+              Some log
+          else
+              None
+
+      // Use the old API on Mono, with ToolsVersion = 12.0
+      let CrackProjectUsingOldBuildAPI(fsprojFile:string) = 
+          let engine = new Microsoft.Build.BuildEngine.Engine()
+          try
+             engine.DefaultToolsVersion <- "12.0"
```
7sharp9: Setting this to 12.0 here means that version 4.0 msbuild tasks will throw a invalid cast exception, we have had to maintain a local copy of FCS with this patched.  In addition it also means the version 4.0 msbuild task will only work in mono throws an exception when this is set to 12.0 which it doesn't do in the latest version.

dsyme:
@rneatherway  Very clear explanation of the choices.  Yes, option 3 is better than option 2.  You can leave the removed methods in the FCS API marked as Obsolete with IsError=true, and a message saying that you can now find them in the new nuget package,

src/fsharp/FSharp.Compiler.Service.ProjectCracker/Program.fs
rneatherway: > Setting this to 12.0 here means that version 4.0 msbuild tasks will throw a invalid cast exception

Ah, I thought the assembly binding redirect would address that?

I originally set it to 12.0 because otherwise mono msbuild didn't work correctly on projects with ToolsVersion set to 12.0 (hence the test with the same name in this repo). Is this fixed?

7sharp9: If its set to 4.0 then the assembly redirects take care of the correct
version loading, it you set it to 12.0 then your moving the default tools
version too high for those tasks which dont explicitly set one.

rneatherway: OK, thanks for the info, will incorporate this change.

7sharp9:
Would we be able to get a new release of FCS that either has no project cracking or reverts the tools version change back to 4.0?  We haven't had a working nuget package in more than a month now :-(

rneatherway:
I'm working on it, but I don't have a lot of time in the evenings. My change does remove the project cracking from FCS.dll. Which commit are you proposing to revert?

7sharp9:
I was referring to this one: https://github.com/7sharp9/FSharp.Compiler.Service/commit/66cf1d24bced4b9bc22a53db200aa7827c67f4a3

7sharp9:
Removing the cracking and releasing as is would be fine by me as we already do the cracking separately now.

rneatherway:
Reverting that would break the cracking on Mono 4.0 wouldn't it? How does this affect you if you already do the cracking separately?

rneatherway:
@7sharp9 2.0.0.0-beta is up on NuGet now, hope that works for you.

Issue #469, opened 2015-11-27
Meaningful exception message needed from failed expression evaluation


andrew-vandenbrink:
```
var session = Create(fsiExePath);
var option = session.EvalExpression(script);
```
An exception of type 'System.Exception' occurred in FSharp.Compiler.Service.dll but was not handled in user code

{"Evaluation failed"}

This exception message is meaningless and we need to be able to diagnose what is wrong with the script instead of receiving message "Evaluation failed"

7sharp9:
Errors would be sent to the output writer.

andrew-vandenbrink:
That should be the case if my script write the error manually to the output

dsyme:
Fixed by https://github.com/fsharp/FSharp.Compiler.Service/pull/475.  When the new nuget is published please use ``session.EvalExpressionNonThrowing(script)``

Issue #161, opened 2014-06-17
Get the exception that caused the evaluation to fail


forki:
Consider the evalExpression function as describe at http://fsharp.github.io/FSharp.Compiler.Service/interactive.html

no try to evaluate a failwith expression:

```
evalExpression "failwith \"inception\" "
```

this causes the session to fail (which is expected). But how can I get the raised exception?

PR #475, opened 2015-12-02
address 469 - capture proper FSharpErrorInfo and exceptions from evaluation

dsyme:
This addresses https://github.com/fsharp/FSharp.Compiler.Service/issues/469 by adding ``EvalInteractionNonThrowing`` and friends that return two things

- an FSharpErrorInfo array for warnings and errors during type checking and parsing
- a choice between a "good" result and a single exception arising from evaluation.


dsyme:
Also addresses issues like https://github.com/fsharp/FSharp.Compiler.Service/issues/161

PR #476, opened 2015-12-02
trial a travis build matrix (Linux, OSX and Mono latest/4.0.5)

dsyme:


dsyme:
OK, there are failures on Mac (Mono master) and also both Mac and Linux for Mono 4.0.5.

PR #478, opened 2015-12-02
Project cracking fix child references

### LINKED ISSUE #477: Failures on Mac (Mono master) and Max/Linux (Mono 4.05)
I've enabled a build matrix in #476 but there are failures

    https://travis-ci.org/fsharp/FSharp.Compiler.Service/jobs/94301490

@rneatherway If you get a chance to look that would be awesome.  The Mac+mono master  failure is in a recently adjusted part of the code

rneatherway:
Should fix #477, depends on #476

### DIFF
--- a/src/fsharp/FSharp.Compiler.Service.ProjectCracker.Exe/Program.fs
+++ b/src/fsharp/FSharp.Compiler.Service.ProjectCracker.Exe/Program.fs
@@ -378,20 +378,23 @@ module Program =
       // C# referenced projects, as we don't look at them otherwise.
       let referencedProjectOutputs =
           if runningOnMono then
-              [| yield! Array.map (fun (s,_) -> "-r:" + s) referencedProjectOptions
-                 for file in parsedProject.ProjectReferences do
-                     let ext = Path.GetExtension(file)
-                     if ext = ".csproj" || ext = ".vbproj" then
-                         let parsedProject = FSharpProjectFileInfo.Parse(file, properties=properties, enableLogging=false)
-                         match parsedProject.OutputFile with
-                         | None -> ()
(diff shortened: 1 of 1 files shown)


src/fsharp/FSharp.Compiler.Service.ProjectCracker.Exe/Program.fs
```diff
@@ -378,20 +378,24 @@ module Program =
...
+                        match parsedProject.OutputFile with
+                        | None -> ()
+                        | Some f -> yield "-r:" + f ]
           else
-              [||]
+              []
+
+      // On some versions of Mono the referenced projects are already
+      // correctly included, so we make sure there aren't any duplicates
+      let cmdOptions = Array.ofSeq (Set.union (set parsedProject.Options)
```
dsyme: Hmm..  I'd be more comfortable if 

1. keep the options in the same order and 

2. we only remove duplicates for ``-r`` options.  I'm concerned that sequences like ``--optimize-`` ``--optimize+`` ``--optimize-`` will just become ``--optimize-`` ``--optimize+`` - I'm not sure this can really happen but it's best to avoid the possibility.

dsyme:
One comment - if you can fix that then we can merge. Great to see things passing.

src/fsharp/FSharp.Compiler.Service.ProjectCracker.Exe/Program.fs
rneatherway: Very good point, adjusted significantly.

rneatherway:
OK, all good.

dsyme:
Great!

rneatherway:
@dsyme should we go ahead and publish a new NuGet package then?

PR #484, opened 2015-12-03
Use JSON rather than binary serialization

rneatherway:
This change also renames ProjectCracker.Tool to ProjectCrackerTool.
This avoids a name clash between type and module on
SourceCodeServices.ProjectCracker, which would otherwise occur
due to the shared source file.


dsyme:
Great! :)

rneatherway:
OK all that is left is to double check what actually happens when referencing the NuGet package. I can try this with FsAutoComplete today.

7sharp9:
Is this ready for nuget release now?

rneatherway:
Yeah, it does seem to work. But in FsAutoComplete I added the NuGet reference using Paket to the FsAutoComplete.Core project, which added both ProjectCracker.dll and ProjectCrackerTool.exe as referenced so they end up in FsAutoComplete.Core/bin/Debug. So far so good, but then only ProjectCracker.dll and *not* ProjectCrackerTool.exe ends up in FsAutoComplete/bin/Debug, so I have to reference ProjectCrackerTool.exe from FsAutoComplete although it doesn't use it directly.

rneatherway:
Right, so as far as I can tell it's because FsAutoComplete.Core doesn't *actually* depend on ProjectCrackerTool.exe, even though it references it, and therefore MSBuild won't copy it over when building FsAutoComplete because it has no way of knowing that it is actually needed.

However, the NuGet package is missing the config file so that needs fixing.

PR #485, opened 2015-12-04
Include ProjectCrackerTool.exe.config in NuGet pkg

rneatherway:


dsyme:
I wonder if you can add ProjetCrackerTool.exe and ProjectCrackerTool.exe.config as linked resources to ProjectCracker.dll by using command line flags. Having them as linked resources should mean that they form part of a multi-file assembly  - it's a group of files but logically .NET and MSBuild should consider them together as one thing.

dsyme:
However I don't know if nuget respects this....

rneatherway:
I tried: <OtherFlags>--linkresource:$(OutputPath)\FSharp.Compiler.Service.ProjectCrackerTool.exe</OtherFlags>

but it doesn't seem to have had the intended effect.

So far I am using:

```
  <Target Name="AfterBuild">
     <Copy SourceFiles="..\packages\FSharp.Compiler.Service.ProjectCracker\lib\net45\FSharp.Compiler.Service.ProjectCrackerTool.exe" DestinationFolder="$(OutputPath)" />
     <Copy SourceFiles="..\packages\FSharp.Compiler.Service.ProjectCracker\lib\net45\FSharp.Compiler.Service.ProjectCrackerTool.exe.config" DestinationFolder="$(OutputPath)" />
  </Target>
```

dsyme:
> but it doesn't seem to have had the intended effect.

Could you elaborate?  Meaning that add-nuget-package ignored the linked resource?  (In retrospect I would expect you to still have to copy the item as you're doing )

thx

rneatherway:
The intended effect was that referencing ProjectCracker.dll in a project should cause the ProjectCrackerTool.exe and .config to be copied to the output path as well. I wasn't sure how to verify that the resource was correctly linked. I didn't use `add-nuget-package` (a Visual Studio thing?), but rather Paket. Things work fine with the Copy, so I will just make sure the documentation is clear about the need to do that.

dsyme:
Yes I was wondering about the behaviour of Visual Studio with an add-nuget-package from the menus then the DLLs in the package contain linked resources.

(If Visual Studio respects the linked resources (e.g. by adding CopyFile's explicitly) then I suppose Paket should to)

The Copy is ok for now though!  Getting some docs in place is necessary either way

PR #503, opened 2015-12-26
Expose QualifiedName and FileName of FSharpImplementationFileContents

alfonsogarciacaro:
This is just to expose the `QualifiedName` and `FileName` properties of `FSharpImplementationFileContents` (I'm mostly interested in the second one). Please disregard if there's any reason to keep these members hidden.


dsyme:
Released in 2.0.0.2

alfonsogarciacaro:
Thanks for that!

PR #519, opened 2016-01-26
Change nuget layout for ProjectCracker package

rneatherway:
Not sure this is right yet, hoping that this PR will trigger the AppVeyor nuget package builder so I can test installing it from there.


rneatherway:
OK, I've tested this using the appveyor nuget feed and it works. Ready for review.

PR #530, opened 2016-02-10
Adjust ProjectCracker NuGet for VS/NuGet

rneatherway:
The Target name cannot include '.', and the .targets file must match the
name of the NuGet package.

### DIFF
--- a/nuget/projectcracker.template
+++ b/nuget/projectcracker.template
@@ -12,7 +12,7 @@ iconurl https://raw.github.com/fsharp/FSharp.Compiler.Service/master/misc/logo.p
 tags
     F#, fsharp, msbuild, editor
 files
-    ../src/fsharp/FSharp.Compiler.Service.ProjectCrackerTool/FSharp.Compiler.Service.ProjectCrackerTool.targets ==> build/net45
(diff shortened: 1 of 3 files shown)


rneatherway:
@cloudroutine is it possible for you to try installing the test packages at https://ci.appveyor.com/project/rneatherway/fsharp-compiler-service/build/artifacts ?

cloudRoutine:
How do I go about doing this? I tried downloading the packages into a directory and adding the directory as a local nuget source inside of VS but when I try to install it says it can't resolve all of the dependencies on my machine and fails.

I tried doing it with paket adding the directory as a local source, but it kept pulling the release version off of nuget instead.

I'm not really sure how to do this.  Another option is to put a prerelease/beta version on nuget; that'd be much simpler to work with.

rneatherway:
@dysme OK with me publishing a prerelease version from this branch?

rneatherway:
I found that you can do it locally this way: https://stackoverflow.com/questions/10240029/how-to-install-a-nuget-package-nupkg-file-locally

Anyway, it works so I'll release.

PR #540, opened 2016-03-10
Fix #539

### LINKED ISSUE #539: BasicPatterns.NewDelegate shows same value for different arguments
Please find below a sample method and the generated AST for its body:

```
let makeDelegate f = System.Func<int,int,int> f

NewDelegate
  (type System.Func<Microsoft.FSharp.Core.int,Microsoft.FSharp.Core.int,Microsoft.FSharp.Core.int>,
   Lambda
  (val delegateArg,
   Lambda
  (val delegateArg,
   Application (Value val f,[],[Value val delegateArg; Value val delegateArg]))))
```

The problem here is that the two automatically generated arguments use the same value `delegateArg`. In other cases of automatically generated closures, arguments are created correctly with different names. 

Now it's time to cry for a little help: whenever I try to compile somewhat complex code I'm finding cases like this one which makes the AST not 100% reliable to compile F# to other platforms. This is a pity because having access to the typed AST it's otherwise a great feature. I would really appreciate if I could get some feedback on these issues: what to expect about the difficulty to fix them, some guidance on how to identify them myself...

Thanks a lot in advance!

alfonsogarciacaro:
WIP. Just a quick shoot at fixing #539. I'll test it later. Comments are welcome.

### DIFF
--- a/src/fsharp/TypeRelations.fs
+++ b/src/fsharp/TypeRelations.fs
@@ -2249,7 +2249,7 @@ let BuildNewDelegateExpr (eventInfoOpt:EventInfo option, g, amap, delegateTy, in
             if List.exists (isByrefTy g) delArgTys then
                     error(Error(FSComp.SR.tcFunctionRequiresExplicitLambda(List.length delArgTys),m)) 
 
-            let delArgVals = delArgTys |> List.map (fun argty -> fst (mkCompGenLocal m "delegateArg" argty)) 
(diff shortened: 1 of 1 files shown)


dsyme:
This fix is fine.  We should add a test too, though I'll merge as is.

dsyme:
Fix is in 2.0.0.7

Issue #538, opened 2016-03-06
BackgroundCompiler.ParseAndCheckProjectImpl takes a very long time on a big solution with a very connected project dependency graph


matzoliv:
I tried to use the "Find all references" feature of F# power tools on a big solution. In certain cases, the analysis would was very slow (I did not measure how slow, I killed devenv.exe after one hour). A quick  profiling showed that the FSharp.Compiler.Service spent all its time running IncrementalBuilder.GetLogicalTimeStampForProject. I then tried to reproduce the degenerative behaviour on a simple example and came up with this.

- A simple solution with 4 projects A, B, C and D.
- B references A
- C references A and B
- D references A, B, and C

Using F# powertools, finding all references on a type defined in A results in :
- 53 calls of GetLogicalTimeStampForProject on the project A
- 21 calls of GetLogicalTimeStampForProject on the project B
- 7 calls of  GetLogicalTimeStampForProject on the project C

I understand that the logical timestamp of a project is computed from the last modified timestamp on all of its files, and recursively from the logical timestamp of all the projects it references. So I believe we're faced with an algorithm that is doing IO with exponential complexity.

In my opinion, simply memoizing the logical timestamps of projects during the computation should solve the problem. I may get enough time in the next weeks to try to do it myself but I thought I would share my analysis in the mean time.

vasily-kirichenko:
I think this is the same issue as in https://github.com/fsprojects/VisualFSharpPowerTools/issues/1280

dsyme:
@vasily-kirichenko @matzoliv Can you verify if https://github.com/fsharp/FSharp.Compiler.Service/pull/550 has reduced the calls to GetLogicalTimeStampForProject? Thanks

@vasily-kirichenko I'm not sure this is the same as https://github.com/fsprojects/VisualFSharpPowerTools/issues/1280 but perhaps it is

dsyme:
Fix is in 2.0.0.8

vasily-kirichenko:
I does not seem to make any significant change. Moving the cursor from one symbol to another causes ~1200 file accesses (VFPT solution). `SafeExists` and `GetLastWriteTimeShim` are always at the top in profiler:

![1](https://cloud.githubusercontent.com/assets/873919/14173459/a706e6b4-f746-11e5-8171-b05683e0a558.gif)

That said, the performance is adequate for editing small to middle size files and awful for "Find all references" on large solutions.

dsyme:
[...]

PR #550, opened 2016-03-31
add time stamp cache

dsyme:
This should fix https://github.com/fsharp/FSharp.Compiler.Service/issues/538 but this needs to be verified.

A TimeStampCache is added for the duration of a background build or other single request for results

Documentation is also added for the ``DependencyFiles`` property and when it can be used.  The DependencyFiles resulting from checking individual files are also reduced to a more appropriate set.


src/fsharp/vs/IncrementalBuild.fs
```diff
@@ -1163,6 +1166,26 @@ type IncrementalBuilder(frameworkTcImportsCache: FrameworkImportsCache, tcConfig
         let flags = tcConfig.ComputeCanContainEntryPoint(sourceFiles |> List.map snd)
         (sourceFiles,flags) ||> List.map2 (fun (m,nm) flag -> (m,nm,flag))
 
+    let defaultTimeStamp = DateTime.Now
+    let getFileTimeStamp (cache: TimeStampCache) fileName = 
+        let ok, v = cache.Files.TryGetValue(fileName)
```
vasily-kirichenko: Just curious: why do you not use `match cache.Files.TryGetValue(fileName) with true, v -> ...`? Is your code more performant?

src/fsharp/CompileOps.fs
```diff
@@ -1749,13 +1749,21 @@ type IRawFSharpAssemblyData =
     abstract HasAnyFSharpSignatureDataAttribute : ILGlobals -> bool
     abstract HasMatchingFSharpSignatureDataAttribute : ILGlobals -> bool
 
-type IProjectReference = 
+type TimeStampCache() = 
+    let cacheCreation = DateTime.Now
```
catlion: Sorry for being late, but using `DateTime.UtcNow` everywhere not only a faster way, but mitigates UTC adjustments problems https://blogs.msdn.microsoft.com/kirillosenkov/2012/01/10/datetime-utcnow-is-generally-preferable-to-datetime-now/

src/fsharp/vs/IncrementalBuild.fs
```diff
@@ -1163,6 +1166,26 @@ type IncrementalBuilder(frameworkTcImportsCache: FrameworkImportsCache, tcConfig
         let flags = tcConfig.ComputeCanContainEntryPoint(sourceFiles |> List.map snd)
         (sourceFiles,flags) ||> List.map2 (fun (m,nm) flag -> (m,nm,flag))
 
+    let defaultTimeStamp = DateTime.Now
```
catlion: Sorry for being late, but using DateTime.UtcNow everywhere not only a faster way, but mitigates UTC adjustments problems https://blogs.msdn.microsoft.com/kirillosenkov/2012/01/10/datetime-utcnow-is-generally-preferable-to-datetime-now/

PR #553, opened 2016-04-11
add Interactive Settings assembly to nuget package

### LINKED ISSUE #552: Nuget for FSharp.Compiler.Interactive.Settings.dll?
Would it be ok to create a Nuget package for this assembly?

It would make things a lot simpler for me if I can directly reference the dll from a known location on all platforms. This would be to workaround the [useFsiAux bug](https://github.com/fsharp/FSharp.Compiler.Service/issues/127#issuecomment-208392104)

nosami:
fixes #552

### DIFF
--- a/nuget/paket.template
+++ b/nuget/paket.template
@@ -19,7 +19,9 @@ files
     ../bin/v4.0/FSharp.Compiler.Service.xml ==> lib/net40
     ../bin/v4.0/FSharp.Compiler.Service.?db ==> lib/net40
     ../bin/v4.0/FSharp.Compiler.Service.dll.?db ==> lib/net40
+    ../bin/v4.0/FSharp.Compiler.Interactive.Settings.dll ==> lib/net40
     ../bin/v4.5/FSharp.Compiler.Service.dll ==> lib/net45
     ../bin/v4.5/FSharp.Compiler.Service.xml ==> lib/net45
     ../bin/v4.5/FSharp.Compiler.Service.?db ==> lib/net45
     ../bin/v4.5/FSharp.Compiler.Service.dll.?db ==> lib/net45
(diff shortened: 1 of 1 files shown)


7sharp9:
We really need this, having the file missing makes this really broken in the current code as theres no way to use the API without the missing file.

dsyme:
OK.  But we will almost certainly move it out into a separate nuget package at some point, once the bug that requires the presence of this DLL at all actually gets fixed.

dsyme:
Actually, I think we can't do this.  We can't have every user of FSharp.Compiler.Service suddenly shipping their own FSharp.Compiler.Interactive.Settings.dll.  Let me take a look at the issue that's causing this to be required.

nosami:
In that case, should I just make it a separate package now then?

dsyme:
@nosami Yes, I suppose so.  I suppose really we should make it in http://github.com/fsharp/fsharp, since it's really one of the components of the F# programming model.  

FCS is _intended_ to not require any specific implementation of the ``fsi`` object. but rather to work with whatever is given as long as it has the right API. But for some reason a reference is still being required even when you pass in ``useFsiAux=false``

nosami:
In my case, I need it to behave as much like fsharpi as possible

dsyme:
@nosami OK, so you want to reference FSharp.Compiler.Interactive.Settings.dll, that's fair enough

Yes, please add a nuget package to http://github.com/fsharp/fsharp.  Push the first edition(s) of the package manually, no need to wait for the PR.  Put ``fsgit`` as an owner of the package in addition to yourself.

dsyme:
@nosami Do you still need this?  I think we should close this otherwise

nosami:
Just saw this, sorry. I forgot all about this. I worked around the issue
for now by adding the missing DLL to source control. I should fix this, but
the priority is low.

Issue #535, opened 2016-02-25
Necessary references to retrieve AST from script files (fsx)


alfonsogarciacaro:
According to the [docs](https://fsharp.github.io/FSharp.Compiler.Service/typedtree.html), in order to get the AST from a F# script file, we must use `FSharpChecker.GetProjectOptionsFromScript` first. This method adds many references by default to the project options (see image) and there's been reports that these references fail in some machines (see https://github.com/fsprojects/Fable/issues/35). I've tried removing these "core" references and the AST can still be retrieved. So te question is: is it safe to remove these references when we only want to retrieve the AST? And also, is there an easy way to distinguish core references from custom ones? (those set explicitly by the user with `#r "..."`)

<img width="867" alt="screen shot 2016-02-25 at 14 50 13" src="https://cloud.githubusercontent.com/assets/8275461/13321789/725b4814-dbd1-11e5-9175-d93219dd91a5.png">

alfonsogarciacaro:
I have closed the issue in Fable as it's probably better to update F# in the failing machines that messing around with the script references, so this is not necessary any longer.

davidpodhola:
Note unfortunately this applies to the final application too see https://github.com/fsprojects/Fable/issues/58. I think the final application should work without F# installed.

alfonsogarciacaro:
That would be ideal, yes. But probably we need to wait unit there's a CoreCLR version of the compiler services.

PR #561, opened 2016-05-02
Dotnet Core tests - wip

ncave:
Initial setup, work in progress. Some of the tests (but not all) are failing on dotnet core being different, others are not yet feasible due to dependency on ProjectCracker, etc.
Comments / advise is welcome.


tests/FSharp.Compiler.Service.Tests.netcore/project.json
```diff
@@ -0,0 +1,70 @@
...
+    "../service/InteractiveCheckerTests.fs",        
+    
+    "Program.fs",
+  ],
+  "compileFiles_not_included_yet": [
+    "../service/FsiTests.fs",
+    "../service/FscTests.fs",
+    "../service/ProjectOptionsTests.fs",
+    "../service/PerfTests.fs",
+    "../service/ExprTests.fs",
```
dsyme: Of these, I think ``ExprTests`` should likely pass relatively easily?

dsyme:
This looks really good, and great to see it is green already (although I'm not sure if netcore build + test is running in CI yet?)

ncave:
@dsyme No, not running the netcore test target in CI yet, it needs to be added as a different platform in the appveyor matrix, so it doesn't impact the other tests.

ncave:
@dsyme I added the dotnet core tests to the build matrix in appveyor so we can see which tests are failing. The dotnet core test target is currently ignored if it fails, as there are a lot of tests that still need to be fixed. I couldn't find a way to show the individual build job status separately (on the same branch), but you can click on the latest appveyor build and see status there. If anything catches your eye on a particular failing test as an easy fix, any advice will be highly appreciated. I don't have a working debug setup (still waiting for rc2 tooling) and the only way to debug right now is with print statements, so progress is slow. If you have a better debug setup, please share.

ncave:
@dsyme All included tests now passing on netcore RC2.

src/fsharp/FSharp.Compiler.Service.ProjectCracker.netcore/ProjectCrackerTool.fs
```diff
@@ -0,0 +1,361 @@
+﻿namespace Microsoft.FSharp.Compiler.SourceCodeServices.ProjectCrackerTool
```
dsyme: This looks more or less duplicated from https://github.com/fsharp/FSharp.Compiler.Service/blob/340b9ab6ca1dd90ab7fcd89d503acc4553a3bca9/src/fsharp/FSharp.Compiler.Service.ProjectCrackerTool/Program.fs.  Do you think we can avoid that duplication?

dsyme:
This looks great.  

I suppose there may be further change in this area ahead given this announcement https://blogs.msdn.microsoft.com/dotnet/2016/05/27/making-it-easier-to-port-to-net-core/ and the other changes in project.json.    But this looks right at this stage.  For example, I suspect we might be able to use MSBuild and paket for the build in due course given the drift back to the XML-based project file tooling.  But we can deal with that in another iteration.

Do you think you could update this with the necessary changes to make sure the FSharp.Compiler.Service.dll is included in the NuGet package (presumably as a .NETStandard 1.5 profile component?)

dsyme:
(If you want to do the NuGet package additions in a separate PR that's also OK, I think we can merge this as is)

src/fsharp/FSharp.Compiler.Service.ProjectCracker.netcore/ProjectCrackerTool.fs
ncave: This duplication is intentional to minimize the impact on other projects, as it is a small refactoring (breaking the above mentioned Program.fs in two files so it can be reused), but yes, I can definitely do that, either here or in a subse

ncave:
Another PR would probably be better, to get some feedback on the versioning. The build already copies the nuget packages into the top bin folder.

ncave:
Also to be addressed in a subsequent PR (I'll open an issue), there is still an outstanding problem with checker.GetProjectOptionsFromScript(file, code), similar to [#535](https://github.com/fsharp/FSharp.Compiler.Service/issues/535) and [#567](https://github.com/fsharp/FSharp.Compiler.Service/issues/567), that probably needs to be fixed before pushing an official netcore version to NuGet. Tests are currently avoiding it (netcore only) by not using it:
```
#if TODO_REWORK_ASSEMBLY_LOAD
    let dllName = Path.ChangeExtension(file, ".dll")
    let projName = Path.ChangeExtension(file, ".fsproj")
    let args = mkProjectCommandLineArgs (dllName, [file])
    let projectOptions = checker.GetProjectOptionsFromCommandLineArgs (projName, args)
#else    
    let projectOptions = checker.GetProjectOptionsFromScript(file, input) |> Async.RunSynchronously
#endif
```

dsyme:
OK, I will merge this then.

PR #562, opened 2016-05-06
add FSharpEntity.ArrayRank

Jand42:
Simply exposing `rankOfArrayTyconRef`. Fails if entity is not an array.


dsyme:
Hi @Jand42 , Could you add a test too please?  Thanks

Jand42:
Currently I am using `td.DisplayName.Length - 1` as a workaround, a bit hackish, but works as proposed `td.ArrayRank`.

dsyme:
Great, thanks for the contribution!

PR #564, opened 2016-05-12
Fixed empty symbol declared pdb.

kekyo:
* (Pre discover input F# source code and got AST.)
* I'm applying "pdbFile" argument in SimpleSourceCodeServices.Compile function with AST, created pdb file specified path, but no symbol entries containing in pdb file.
* The "pdbPath" handled main1OfAst function, but no affected tcConfigB and tcConfig debuginfo property.
* PR contains if pdbPath is applied, set debuginfo is true by SetDebugSwitch.

### DIFF
--- a/src/fsharp/fsc.fs
+++ b/src/fsharp/fsc.fs
@@ -1961,7 +1961,10 @@ let main1OfAst (openBinariesInMemory, assemblyName, target, outfile, pdbFile, dl
     tcConfigB.framework <- not noframework 
     // Preset: --optimize+ -g --tailcalls+ (see 4505)
     SetOptimizeSwitch tcConfigB OptionSwitch.On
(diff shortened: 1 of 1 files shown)


dsyme:
Does it work if you pass a ``--debug+`` argument explicitly?

kekyo:
Thx reply, I'm trying with add "--debug+" option, but symbol empty too.

Test codes:
  https://gist.github.com/kekyo/a39cac521df12d6c9d762af3e30c0bd6

kekyo:
@dsyme Any idea?

dsyme:
@kekyo Could you edit the issue so the steps are using [ISSUE_TEMPLATE.md](https://github.com/fsharp/FSharp.Compiler.Service/blob/master/ISSUE_TEMPLATE.md) format please?

kekyo:
Thank you and sorry issue report too late...

PR #566, opened 2016-05-14
Project cracker: Set properties before loading the project.

### LINKED ISSUE #565: ProjectCracker fails to detect references
Project Cracker Tool fails to detect references for the attached project. 
[ConsoleApplication3.zip](https://github.com/fsharp/FSharp.Compiler.Service/files/264555/ConsoleApplication3.zip)
#### Repro steps
1. Download the attached sample solution and open it in VS Code / Atom / other
#### Expected behavior

No errors
#### Actual behavior

Error squigglies on `open System.Data` in the editor. No type information for System.Data is available.

When running Project Cracker Tool, this error is shown:

> {ProjectFile =
> "C:\Users\robert\Documents\Visual Studio 2015\Projects\ConsoleApplication3\ConsoleApplication3\ConsoleApplication3.fsproj";
> Options = [||];
> ReferencedProjectOptions = [||];
> LogOutput =
> "Microsoft.Build.Exceptions.InvalidProjectFileException: The value "" of the "Project" attribute in element is invalid. Parameter "path" cannot have zero length.
> at Microsoft.Build.Shared.ProjectErrorUtilities.ThrowInvalidProject(String errorSubCategoryResourceName, IElementLocation elementLocation, String resourceName, Object[] args)
> at Microsoft.Build.Shared.ProjectErrorUtilities.VerifyThrowInvalidProject(Boolean condition, String errorSubCategoryResourceName, IElementLocation elementLocation, String resourceName, Object arg0, Object arg1, Object arg2, Object arg3)
> ...
#### Related information

Originally reported here: https://github.com/fsharp/FsAutoComplete/issues/96
This machine has VS2015 installed, and has never had VS2013 installed. 
- Operating system : Win 10

rojepp:
Fixes #565 

The properties need to be set *before* loading the project. `LoadProject` takes a dictionary for properties, so here we make one and populate the values, instead of setting them afterwards. `LoadProject` would throw when properties were not set.

### DIFF
--- a/src/fsharp/FSharp.Compiler.Service.ProjectCrackerTool/Program.fs
+++ b/src/fsharp/FSharp.Compiler.Service.ProjectCrackerTool/Program.fs
@@ -139,17 +139,15 @@ module internal Program =
           let projectInstanceFromFullPath (fsprojFullPath: string) =
               use stream = new IO.StreamReader(fsprojFullPath)
               use xmlReader = System.Xml.XmlReader.Create(stream)
-
-              let project = engine.LoadProject(xmlReader, FullPath=fsprojFullPath)
               
-              project.SetGlobalProperty("BuildingInsideVisualStudio", "true") |> ignore
-              if not (List.exists (fun (p,_) -> p = "VisualStudioVersion") properties) then
-                  match vs with
-                  | Some version -> project.SetGlobalProperty("VisualStudioVersion", version) |> ignore
-                  | None -> ()
-              project.SetGlobalProperty("ShouldUnsetParentConfigurationAndPlatform", "false") |> ignore
-              for (prop, value) in properties do
-                    project.SetGlobalProperty(prop, value) |> ignore
+              let props = System.Collections.Generic.Dictionary()
+              for (k,v) in properties do 
+                 props.Add (k, v)
+              if not (props.ContainsKey "VisualStudioVersion") then do
+                   vs |> Option.iter (fun v -> props.Add ("VisualStudioVersion", v))
+              props.Add ("BuildingInsideVisualStudio", "true")
+              props.Add ("ShouldUnsetParentConfigurationAndPlatform", "false")
+              let project = engine.LoadProject(xmlReader, props, engine.DefaultToolsVersion, FullPath=fsprojFullPath)
 
               project.CreateProjectInstance()
 


rneatherway:
Thanks very much for tracking this issue down! It's hard to reproduce some of these edge cases without having exactly the same machine configuration. Unfortunately this PR breaks the MSBuild v12 case. One option would be for the redirect function (`addMSBuildv14BackupResolution`) to store whether that redirect needed to be applied and then to use that fact to determine whether to choose your new code path or the existing one. It doesn't feel that clean, but then we just want things to work!

dsyme:
@rneatherway Should we update and accept this?

rneatherway:
Do you mean with my suggested adjustment in the comment above? I should be able to find time for that this weekend. Hopefully at some point we can just drop support for VS2013, or require the installation of MSBuild v14 if that is available separately.

7sharp9:
Could an alternative cracker be used here to remove the msbuild dependency?

When I worked for Xamarin we worked to remove the project cracker because it was far too slow, and it was also preferable to not have external process invoke which is required by the use of msbuild.

rneatherway:
Well, that would theoretically be great, but at the end of the day the references are essentially determined  by MSBuild and I can't imagine someone writing a robust MSBuild simulator. I also don't find the speed to be an issue particularly, it's just a few hundred ms on project load.

7sharp9:
I wish I get a few hundred ms, to be honest I was lucky to get results in 4-6 seconds.  I don't honestly think you need a simulator, the logic to load assemblies is not too difficult, the only complexity is pcl, ios and android which all behave similarly with a directory full of dlls tagged onto the end of the assembly list.

rneatherway:
Wow, I've never seen that kind of speed.

You'd have to write a parser for fsproj and csproj files that understood all the different Choice tags and things like that. It doesn't seem that easy to me, but I would be happy to give something a go if you got around to it.

7sharp9:
@rneatherway Thats why we removed it from Xamarin studio :-)

dsyme:
> to be honest I was lucky to get results in 4-6 seconds. 

It would be good to have an issue (with repro) recorded for this kind of problem, rather than leaving it as anecdotal.   There may be something very simple that could be done to fix this.

7sharp9:
I would if I could, I don't use mono anymore.

dsyme:
@rojepp Could you update this by merging with master please?  Thanks

cloudRoutine:
@rojepp any intention of updating this?
@rneatherway  you could also just checkout the PR locally and update it to master yourself so this could be merged, it's an important fix.
@dsyme otherwise I can just make a new PR with the exact same change, but I'd rather not steal the credit from @rojepp

rojepp:
Sorry about the delay everybody. I thought it easier to just make a new PR: https://github.com/fsharp/FSharp.Compiler.Service/pull/626

cloudRoutine:
@rneatherway [msbuild 14 tools are available separately btw](https://www.microsoft.com/en-us/download/details.aspx?id=48159)

PR #572, opened 2016-06-04
Added FSharpTargetsPath condition

ncave:


dsyme:
Why is this needed - won't the build fail if the targets can't be found?

ncave:
@dsyme FSharp.Compiler.Service.fsproj already has that condition, this just brings the rest of the projects in line.
Also, it gives a much better error message instead of an obscure one when msbuild is not there.

PR #574, opened 2016-06-08
Minor test fix for netcore xplat

ncave:
Minor fix to make tests pass on Linux.


dsyme:
So .NET Core FCS tests are running on Linux - that's great, I hadn't quite realized that.

ncave:
@dsyme They do, but the xplat codegen and build tooling is not there yet, so you need to skip it.

Issue #582, opened 2016-06-17
FCS.ProjectCracker on dotnet core - msbuild version


ncave:
Here is an outstanding issue for the dotnet core implementation of FCS.ProjectCracker:

1) The FSharp.Compiler.Service.ProjectCracker depends on building the cracked project with MSBuild to get the references from it, but:
2) The latest MSBuild (installed per [this](https://github.com/Microsoft/msbuild/wiki/Building-Testing-and-Debugging-on-.Net-Core-MSBuild#unofficial-getting-net-core-msbuild-binaries-without-building-the-code)) does not seem to be able to build older .fsproj projects, or at least not out of the box when I tried it.

This will probably improve when the next release of VS 15, please share if there is a MSBuild version that works with older fsharp projects.

In the mean time, some progress can be made by taking in [#566](https://github.com/fsharp/FSharp.Compiler.Service/pull/566) so at least the result after cracking has the project options and files, even if it cannot build properly to get the references (perhaps not everybody needs the references).

It's entirely possible that due to limited information I'm missing the big picture on where msbuild xplat is going, if that's the case please ignore.

7sharp9:
Having no msbuild in the equation would be nice.   msbuild dependent project are a major pain, it doesn't have to be that way.

enricosada:
@ncave We need to ask msbuild repo team, i's open now, can you open an issue there about older fsproj? 
also /cc @cartermp too because it's interesting for visualfsharp i think, so if we can get some help it's good for everyone.

I'll open one about build and package. We need a package usable by preview1 (or preview2, but is the the same rc2 netcore) 
Problem is, it's not a good idea to use dev feed. 
Lot of strange stuff there ( rc2 rc3 ), and doing `-*` for deps may give wrong version or not working with preview1 sdk. 
Really it'is a pain to stay bleeding edge in dev feed, let's avoid it if possible, lost too much time. But now there is a working released preview1 so i am happy to use it and avoid pain

ncave:
Here is the link to the issue on the MSBuild repo: [#709](https://github.com/Microsoft/msbuild/issues/709)
Looks like it's not possible at the moment, but perhaps in VS 15.

ncave:
Closing as no longer seems relevant.

PR #583, opened 2016-06-17
Use netcoresdk preview1

enricosada:
the .NET Core SDK is the preview1 ( latest released from http://dot.net ), was already like that but not is explicit.

fix to use only packages released on nuget.org, no dev packages from dev feeds

fix deps, NETCore.App is for console app, NETStandard.Library is of libraries (NETCore.App depends on NETStandard.Library, so i just removed useless dependencies, be lean)

btw about versions:

- rc2 is ok for .net core (corefx,bcl,coreclr)
- rc3 is oftern (always?) wrong, a dev package. there is not going to be a rc3, directly rtm after rc2
- preview1 is for tools/sdk packages


ncave:
Just for the record, yes, MSBuild 0.1.0-preview-00022 is the latest rc2, but if you actually try to use it, it fails with a missing dependency on ms.tpl.dataflow. MSBuild 0.1.0-preview-00024-* looks better (yes, it's rc3, but at least no missing dependencies), but is not able to build older .fsproj projects either (see [#582](https://github.com/fsharp/FSharp.Compiler.Service/issues/582).

dsyme:
Is there an MSBuild issues with repro entered about this?  Which fsproj files have the problem?

Thanks
Don

enricosada:
@ncave We need to ask msbuild repo team, i's open now, can you open an issue there about older fsproj?

I'll open one about build and package. We need a package usable by preview1 (or preview2, but is the the same rc2 netcore) 
Problem is, it's not a good idea to use dev feed. 
Lot of strange stuff there ( rc2 rc3 ), and doing `-*` for deps may give wrong version or not working with preview1 sdk. 
Really it'is a pain to stay bleeding edge in dev feed, let's avoid it if possible, lost too much time. But now there is a working released preview1 so i am happy to use it and avoid pain

ncave:
I have added an issue on the MSBuild repo: [#709](https://github.com/Microsoft/msbuild/issues/709)

PR #588, opened 2016-06-20
Store useMonoResolution flag

### LINKED ISSUE #581: FCS on dotnet core - assembly resolution
Just starting a discussion on some remaining issues for dotnet core support, please feel free to chime in if you have any insight on the best approach.

- There is an outstanding assembly resolution problem with _checker.GetProjectOptionsFromScript(file, code)_, similar to [#535](https://github.com/fsharp/FSharp.Compiler.Service/issues/535) and [#567](https://github.com/fsharp/FSharp.Compiler.Service/issues/567), that may need to be fixed before pushing an official netcore version to NuGet. Tests are currently avoiding it (netcore only) by not using it (using command line args instead):
```
#if TODO_REWORK_ASSEMBLY_LOAD
    let dllName = Path.ChangeExtension(file, ".dll")
    let projName = Path.ChangeExtension(file, ".fsproj")
    let args = mkProjectCommandLineArgs (dllName, [file])
    let projectOptions = checker.GetProjectOptionsFromCommandLineArgs (projName, args)
#else
    // ===> this does not work properly under DotNet Core:
    let projectOptions = checker.GetProjectOptionsFromScript(file, input) |> Async.RunSynchronously
#endif
```

ncave:
Resolves #581

### DIFF
--- a/src/fsharp/CompileOps.fs
+++ b/src/fsharp/CompileOps.fs
@@ -4849,6 +4849,7 @@ module private ScriptPreprocessClosure =
             | CodeContext.Compilation | CodeContext.Evaluation -> MSBuildResolver.CompileTimeLike
 #endif
         tcConfigB.framework <- false 
+        tcConfigB.useMonoResolution <- useMonoResolution
         // Indicates that there are some references not in BasicReferencesForScriptLoadClosure which should
         // be added conditionally once the relevant version of mscorlib.dll has been detected.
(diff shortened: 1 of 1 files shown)


7sharp9:
I can confirm this fixes fsautocomplete with dotnetcore
Thanks @ncave !!!

dsyme:
Super fix, thanks

PR #589, opened 2016-06-20
Added default script references for .NET Core

ncave:
A quick hack to make script parsing work.

### DIFF
--- a/src/fsharp/CompileOps.fs
+++ b/src/fsharp/CompileOps.fs
@@ -1585,6 +1585,25 @@ let GetFsiLibraryName () = "FSharp.Compiler.Interactive.Settings"
 //    (b) included in environment for files 'orphaned' from a project context
 //            -- for orphaned files (files in VS without a project context)
 //            -- for files given on a command line without --noframework set
+#if TODO_REWORK_ASSEMBLY_LOAD
+
+// .NET Core references
+let DefaultBasicReferencesForOutOfProjectSources = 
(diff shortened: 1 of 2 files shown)


dsyme:
@ncave I'm OK with this as a temporary thing, but perhaps you could submit an issue and alignment PR to Microsoft/visualfsharp too?  Ultimately that's the repo where the Microsoft-supported .NET Core packages for F# will be published from. 

Where possible we should include @KevinRansom in this  toodiscussion, since we need to be on a path to convergence in the specification of the .NET Core F# scripting programming model.

dsyme:
@ncave Also, is there any specific reason why you chose this list of types?  Is it the set needed to support FSharp.Core?

ncave:
@dsyme It's an ad-hoc set of dependencies used to pass most of the script tests in the test suite.
I'll open an issue in Microsoft/visualfsharp to possibly get a better solution by @KevinRansom.

dsyme:
Thanks, please link the issue here

FWIW I think we should also be referencing these assemblies (e.g. System.IO) in .NET Framework F# Interactive scripting.  There have been cases that I've had to reference System.IO manually when referencing PCL or .NET Standard components from F# Interactive.  It would be better if we just auto-referenced all the relevant .NET Framework facade assemblies that cover the new-style PCL/.NETStandard surface area for  "mscorlib", "System" and "System.Core"

ncave:
Added an issue in Microsoft/visualfsharp: [#1272](https://github.com/Microsoft/visualfsharp/issues/1272)

PR #590, opened 2016-06-21
.NET Core ProjectCracker - updated version and dependencies

ncave:
@dsyme @enricosada Rationale for this change: As mentioned before [here](https://github.com/fsharp/FSharp.Compiler.Service/pull/583#issuecomment-226913366), MSBuild 0.1.0-preview-00022 does not work at all, but the latest MSBuild 0.1.0-preview-* partially works (at least it gives you the project options and file names, but no references cause it cannot properly build existing projects). This is a temporary measure to provide interim functionality, references will probably change anyway with netcore RTM.

tl;dr IMO it's prudent to have the ProjectCracker working at least partially, for the sake of clients like Fable that can use the project options and file names but may or may not need the references.


dsyme:
Looks reasonable - @enricosada  can you double check please?

PR #595, opened 2016-06-29
Expose MemberFlags for Expr TraitCall

### LINKED ISSUE #580: Information about BasicPatterns.TraitCall
Related to https://github.com/fsprojects/Fable/issues/193

Fable is not resolving properly [BasicPatterns.TraitCall](https://github.com/fsharp/FSharp.Compiler.Service/blob/340b9ab6ca1dd90ab7fcd89d503acc4553a3bca9/src/fsharp/vs/Exprs.fsi#L206-L207) expressions, basically because I wasn't sure how they should be dealt with. Could you please explain a bit about these expressions? When they appear and how they should be resolved (when they correspond to an instance or static call, etc)? Thanks in advance!

alfonsogarciacaro:
This is an attempt to fix #580.

As mentioned in the issue, I'm keeping the `MemberFlags` when [building `E.TraitCall` from `TOp.TraitCall`](https://github.com/fsharp/FSharp.Compiler.Service/blob/340b9ab6ca1dd90ab7fcd89d503acc4553a3bca9/src/fsharp/vs/Exprs.fs#L709). In order not to break compatibility with `BasicPatterns.TraitCall` signature, I'm creating a new module `DerivedPatterns` (as with Quotations).

Please let me know if this is acceptable or another approach should be preferred instead.

Cheers!

### DIFF
--- a/src/fsharp/vs/Exprs.fsi
+++ b/src/fsharp/vs/Exprs.fsi
@@ -204,9 +204,5 @@ module BasicPatterns =
     val (|ObjectExpr|_|) : FSharpExpr -> (FSharpType * FSharpExpr * FSharpObjectExprOverride list * (FSharpType * FSharpObjectExprOverride list) list) option
 
     /// Matches expressions for an unresolved call to a trait 
-    val (|TraitCall|_|) : FSharpExpr -> (FSharpType list * string * FSharpType list * FSharpType list * FSharpExpr list) option 
+    val (|TraitCall|_|) : FSharpExpr -> (FSharpType list * string * Ast.MemberFlags * FSharpType list * FSharpType list * FSharpExpr list) option 
 
-module DerivedPatterns =
-
-    /// Matches expressions for an unresolved call to a trait with MemberFlags info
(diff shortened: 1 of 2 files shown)


src/fsharp/vs/Exprs.fsi
```diff
@@ -206,4 +206,7 @@ module BasicPatterns =
     /// Matches expressions for an unresolved call to a trait 
     val (|TraitCall|_|) : FSharpExpr -> (FSharpType list * string * FSharpType list * FSharpType list * FSharpExpr list) option 
 
+module DerivedPatterns =
 
+    /// Matches expressions for an unresolved call to a trait with MemberFlags info
+    val (|TraitCallExtended|_|) : FSharpExpr -> (FSharpType list * string * Ast.MemberFlags * FSharpType list * FSharpType list * FSharpExpr list) option
```
dsyme: You can just change TraitCall's signature.  We will bump the major version number

dsyme:
As mentioned in the line comment, breaking compat is ok

alfonsogarciacaro:
Thanks, I did that :+1: This is only for my use case but, as we're changing the signature of `BasicPatterns.TraitCall`, if there's something else worth including (there're other values discarded when building  E.TraitCall  from  TOp.TraitCall), please tell me.

PR #601, opened 2016-07-05
Adjust for changes in AppVeyor (?)

dsyme:
The master build is failing - it looks like there is some difference in the AppVeyor install.


dsyme:
It seems that upgrading to Visual Studio 2015 Update 3 causes MSBuild v12.0 to fail on some F# PCL project files with this strange error:

I updated image

It's possible we need to move to MSBuild v14.0 to get these working again.  I presume this is a regression in MSBuild.

PR #610, opened 2016-07-17
Trial the use of MSBuild resolution on latest Mono

dsyme:


dsyme:
This builds on https://github.com/fsharp/FSharp.Compiler.Service/pull/609

dsyme:
(Still getting a null exception from reference resolution)

PR #614, opened 2016-07-26
Xamarin and MonoDevelop addin

wallymathieu:
It looks like this addin has moved to a separate repository.


7sharp9:
Actually its moved from that location now to monodevelop/external/fsharp

wallymathieu:
https://github.com/fsharp/fsharpbinding/blob/master/README.md ?

7sharp9:
It started out at:
https://github.com/fsharp/fsharpbinding/blob/master/README.md
then moved here
but is now here:
https://github.com/mono/monodevelop/tree/master/main/external/fsharpbinding

Issue #637, opened 2016-09-21
Compiling from a fresh clone results in an error


7sharp9:
Compiling from a fresh clone results in an error
#### Repro steps

Provide the steps required to reproduce the problem
1. Have mono installed
2. Have dotnetcore installed
3. ./build.sh
#### Expected behavior

compiles
#### Actual behavior

```
  1) System.Exception: Command failed with exit code 2
  at FSI_0005.Build+assertExitCodeZero@226.Invoke (System.String message) <0x11513c070 + 0x0002f> in <filename unknown>:0
  at Microsoft.FSharp.Core.PrintfImpl+StringPrintfEnv`1[TResult].Finalize () <0x10bbf6bd0 + 0x00039> in <filename unknown>:0
  at Microsoft.FSharp.Core.PrintfImpl+Final1@224[TState,TResidue,TResult,A].Invoke (Microsoft.FSharp.Core.FSharpFunc`2 env, A a) <0x11513be10 + 0x00095> in <filename unknown>:0
  at Microsoft.FSharp.Core.OptimizedClosures+Invoke@3253[T2,TResult,T1].Invoke (T2 u) <0x11513bde0 + 0x00025> in <filename unknown>:0
  at FSI_0005.Build.assertExitCodeZero (Int32 x) <0x11513ade0 + 0x000c0> in <filename unknown>:0
  at FSI_0005.Build+runCmdIn@227.Invoke (System.String args) <0x11513aca0 + 0x0010b> in <filename unknown>:0
  at Microsoft.FSharp.Core.PrintfImpl+StringPrintfEnv`1[TResult].Finalize () <0x10bbf6bd0 + 0x00039> in <filename unknown>:0
  at Microsoft.FSharp.Core.PrintfImpl+Final3@247[TState,TResidue,TResult,A,B,C].Invoke (Microsoft.FSharp.Core.FSharpFunc`2 env, A a, B b, C c) <0x1142e1ac0 + 0x0014f> in <filename unknown>:0
  at Microsoft.FSharp.Core.OptimizedClosures+Invoke@3301-2[T2,T3,T4,TResult,T1].Invoke (T2 u, T3 v, T4 w) <0x1136ea800 + 0x0003d> in <filename unknown>:0
[...]

PR #638, opened 2016-09-21
allow for use of the FCS_DNC env flag to skip doing DNC build steps

baronfel:
should help Dave out some :D

set FCS_DNC to false and then go nuts


7sharp9:
To explain a little more having dnc and mono installed results in a non building FCS. See: #637

dsyme:
@baronfel Thanks for this!

Issue #516, opened 2016-01-20
[Feature Request] Add Compile* methods from SimpleSourceCodeServices to SourceCodeServices to unify API


7sharp9:
#362 Discusses this

Essentially Compile and CompileToDynamicAssembly methods should be brought over from SimpleSourceCodeService to SourceCodeServices.

@dsyme In #362 you mention arguments should be textual to conform with the rest of the API, which methods were you referring to?

Also to provision a PR to add this feature would you be happy with adding the compile helpers to a module within Services.fs so that SimpleSourceCodeServices and SourceCodeServices share the common implementation?

dsyme:
Closing as we don't need to track this with an issue.  We just need to do it :)

PR #639, opened 2016-09-22
Compiler api harmonise

7sharp9:
This PR addresses the fact the `SimpleSourceCodeServices` has compile specific API's that are not present in `SourceCodeServices`.  

The benefits of this PR is that a single `FSharpChecker` can be created and used to do compile operations, rather than having to move between `SimpleSourceCodeServices` and `SourceCodeServices`.

### DIFF
--- a/src/fsharp/vs/service.fs
+++ b/src/fsharp/vs/service.fs
@@ -23,6 +23,7 @@ open Microsoft.FSharp.Compiler.AbstractIL.Internal.Library
 open Microsoft.FSharp.Compiler.AccessibilityLogic
 open Microsoft.FSharp.Compiler.Ast
 open Microsoft.FSharp.Compiler.CompileOps
+open Microsoft.FSharp.Compiler.Driver
 open Microsoft.FSharp.Compiler.ErrorLogger
 open Microsoft.FSharp.Compiler.Lib
 open Microsoft.FSharp.Compiler.MSBuildResolver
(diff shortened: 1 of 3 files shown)


7sharp9:
Fixes / Implements #516

src/fsharp/vs/service.fs
```diff
@@ -2697,6 +2807,66 @@ type FSharpChecker(projectCacheSize, keepAssemblyContents, keepAllBackgroundReso
 
     member ic.TryGetRecentTypeCheckResultsForFile(filename, options, ?source) = ic.TryGetRecentCheckResultsForFile(filename,options,?source=source)
 
+    member ic.Compile(argv: string[]) =
```
dsyme: I think these operations should all go via the backgroundCompiler and the Reactor thread - that's more obvious now that they are all part of the proper service API.  Could you add that as part of this PR?  Or we could do that as a separate item of work.

7sharp9: I think a separate PR as that would be a change of functionality, Im not entirely sure on the interaction of the compile action with an ongoing background compilation.  Maybe that in itself is a good reason for a reactor lock.

7sharp9:
Do we need to depreciate the compile methods in `SimpleSourceCodeServices`? 

Maybe `SimpleSourceCodeServices` should be depreciated entirely?  The only reason Ive ever used it was to use the `CompileFromAsts` method!

dsyme:
> Maybe SimpleSourceCodeServices should be depreciated entirely? The only reason Ive ever used it was to use the CompileFromAsts method!

Yes, I believe it can now be deprecated (though please double check, maybe there's something dangling).

7sharp9:
Theres quite a few uses around on github, I think the bast bet is to mark it as depreciated for now?

dsyme:
> Theres quite a few uses around on github, I think the bast bet is to mark it as depreciated for now, I can add that to this PR.

Great, thanks

dsyme:
Here's the AppVeyor failure:

```
[00:05:40] C:\Program Files\dotnet\dotnet.exe compile-fsc @C:\projects\fsharp-compiler-service\src\fsharp\FSharp.Compiler.Service\obj\Release\netstandard1.6\dotnet-compile.rsp returned Exit Code 1
[00:05:40] C:\projects\fsharp-compiler-service\src\fsharp\vs\service.fs(2189,31): error FS0039: The namespace or module 'AssemblyBuilder' is not defined
[00:05:40] C:\projects\fsharp-compiler-service\src\fsharp\vs\service.fs(2190,29): error FS0072: Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.
```

7sharp9:
Hmm, dotnetcore issue?

7sharp9:
At the moment I have to disable dotnetcore to get mono to compile.  Maybe #637 needs a closer look to get the CI stable.

dsyme:
@7sharp9 Yes.  Either a missing reference, or changed namespace, or need to #if out the functionality, or something.

dsyme:
@7sharp9 I think ``FX_NO_APP_DOMAINS`` is defined on .NET Core and you may need to open ``System.Reflection.Emit`` namespace to find ``AssemblyBuilder``.

7sharp9:
@dsyme Appveyor is ok now.

dsyme:
Thanks for this @7sharp9

7sharp9:
I'll do  a further PR later to add the depreciate flag in simplesourcecodeservces and the reactor thread.

dsyme:
@7sharp9 Thanks.  Let's aim to nuke SimpleSourceCodeServices and bring it all into FSharpChecker.

PR #650, opened 2016-10-13
Add module values to navigable items

vasily-kirichenko:
It also adds `EnclosingEntityKind` and `IsAbstract` to `FSharpNavigationDeclarationItem` which are needed by FSAC -> Ionide VS Code for filtering declaration used by Code Lens feature.

### DIFF
--- a/src/fsharp/vs/ServiceNavigation.fs
+++ b/src/fsharp/vs/ServiceNavigation.fs
@@ -133,6 +133,8 @@ module NavigationImpl =
                   | _ -> (lid, m)
                 [ createMemberLid(lidShow, kind, icon, unionRanges rangeMerge m) ]
             | SynPat.LongIdent(LongIdentWithDots(lid,_), _,_, _, _, _), _ -> [ createMemberLid(lid, FieldDecl, iIconGroupConstant, unionRanges (List.head lid).idRange m) ]
+            | SynPat.Named (_,ident,_,_,r), _ ->
+                [ createMemberLid([ident], FieldDecl, iIconGroupConstant, r)]
             | _ -> []
         
         // Process a class declaration or F# type declaration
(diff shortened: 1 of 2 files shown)


dsyme:
Looks good.

vasily-kirichenko:
@rneatherway @tpetricek @7sharp9 could you please publish a new NuGet package so we can proceed with FSAC and ionide?

7sharp9:
@vasily-kirichenko Sorry my access was revoked?

dsyme:
@7sharp9 You now have access to the nuget package.  I've pushed 8.0.0. This contains the work to optionally remove the MSBuild dependency for script reference resolution (though Visual F# Power Tools should not, I expect, opt-in to that, since it can always assume MSBuild hosted with appropriate binding redirects)

7sharp9:
@dsyme Is there now a programatic way to list the assembly refs within a project file without msbuild?

7sharp9:
@dsyme To be clear, It would be great if FCS could be self contained and provide a means to call resolve on an assembly ref, or ideally return a list of resolved refs from a fsproj file.

The ability to resolve anything is internal as far as I can see, I can see how it could be exposed a bit like what you were doing here:
https://github.com/fsharp/FSharp.Compiler.Service/pull/649/files#diff-d04bc962d64f4e6adc7210878e11a52dR202

Maybe using the msbuild/simulatedmsbuild as per the environment.

vasily-kirichenko:
@dsyme thanks!

dsyme:
> @dsyme Is there now a programatic way to list the assembly refs within a project file without msbuild?

No, this only covered the use of MSBuild assembly resolution for ``#r`` references in scripts.  So the project cracker is still the only way to crack project files.

> @dsyme It would be great if FCS could be self contained and provide a means to call resolve on an assembly ref

We could expose SimulatedMSBuildRefereceResolver but do we really want to?  I suppose it could be particularly useful for putting together support an alternative project file format?

7sharp9:
@dsyme The reason I asked is it would be useful to me so I don't have to start a child process to retrieve them via msbuild.  It would also be useful for any other tooling that may need that info.  Having `resolve` available in the context of the checker should be ok, feeding it properties like framework version etc would be fine.

PR #654, opened 2016-10-14
Added test for checker.GetProjectOptionsFromScript

ncave:
Re-did this PR to just have the test for checker.GetProjectOptionsFromScript
(test failing on netcore in 10.01).

### DIFF
--- a/tests/service/ProjectAnalysisTests.fs
+++ b/tests/service/ProjectAnalysisTests.fs
@@ -4505,10 +4505,16 @@ module Project35b =
     open System.IO
 
     let fileName1 = Path.ChangeExtension(Path.GetTempFileName(), ".fsx")
+#if DOTNETCORE
     let fileSource1 = """
-#r "System.dll"
+#r "System.Private.CoreLib.dll"
 #r "notexist.dll"
 """
+#else
+    let fileSource1 = """
+#r "notexist.dll"
+"""
+#endif
     File.WriteAllText(fileName1, fileSource1)
     let cleanFileName a = if a = fileName1 then "file1" else "??"
 
@@ -4521,6 +4527,23 @@ module Project35b =
     let options = checker.GetProjectOptionsFromCommandLineArgs (projPath, args2)
 #else    
     let options = checker.GetProjectOptionsFromScript(fileName1, fileSource1) |> Async.RunSynchronously
+#endif
+
+[<Test>]
+let ``Test Project35b whole project errors`` () = 
+#if DOTNETCORE
+    let options = checker.GetProjectOptionsFromScript(Project35b.fileName1, Project35b.fileSource1, assumeDotNetFramework=false) |> Async.RunSynchronously
+#else
+    let options = checker.GetProjectOptionsFromScript(Project35b.fileName1, Project35b.fileSource1) |> Async.RunSynchronously
+#endif
+    let wholeProjectResults = checker.ParseAndCheckProject(options) |> Async.RunSynchronously
+    for e in wholeProjectResults.Errors do 
+        printfn "Project35b error: <<<%s>>>" e.Message
+    wholeProjectResults.Errors.Length
+#if DOTNETCORE
+    |> shouldEqual 1
+#else
+    |> shouldEqual 2
 #endif
 
 [<Test>]
--- a/tests/service/ProjectAnalysisTests.fs
+++ b/tests/service/ProjectAnalysisTests.fs
@@ -4530,7 +4530,7 @@ module Project35b =
 #endif
 
 [<Test>]
-let ``Test Project35b whole project errors`` () = 
+let ``Test Project35b whole project errors`` () =
 #if DOTNETCORE
     let options = checker.GetProjectOptionsFromScript(Project35b.fileName1, Project35b.fileSource1, assumeDotNetFramework=false) |> Async.RunSynchronously
 #else
--- a/global.json
+++ b/global.json
@@ -1,3 +1,6 @@
 {
-  "projects": [ "src/fsharp", "tests" ]
+  "projects": [ "src/fsharp", "tests" ],
+  "sdk": {
+    "version": "1.0.0-preview2-003131"
+  }
 }


src/fsharp/CompileOps.fs
```diff
@@ -1643,6 +1643,7 @@ let GetFsiLibraryName () = "FSharp.Compiler.Interactive.Settings"
 // .NET Core references
 let DefaultBasicReferencesForOutOfProjectSources = 
     [   yield Path.Combine(Path.GetDirectoryName(typeof<System.Object>.Assembly.Location),"mscorlib.dll"); // mscorlib
+        yield typeof<System.Object>.Assembly.Location; // System.Private.CoreLib
```
ncave: Yes, we were not testing this particular scenario on netcore. I added a test for it, here is the error if the reference to `System.Private.CoreLib` is not there:
```
Project35b error: <<<The type 'Object' is required here and is unavailable. You must add a reference to assembly 'System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e'.>>>
Project35b error: <<<A reference to the type 'System.Object' in assembly 'mscorlib' was found, but the type could not be found in that assembly>>>
```

ncave:
@dsyme Rebased to FCS latest.

ncave:
@dsyme Here is the same [problem](https://github.com/dotnet/roslyn/issues/14614), same solution at the end.

dsyme:
@ncave Do we expect this to be failing? thanks
```
[00:15:47] 1) Failed : Tests.Service.ProjectAnalysisTests.Test Project35b whole project errors
[00:15:47]   Expected: 2\nActual: 3
[00:15:47]   Expected: 2
[00:15:47]   But was:  3
[00:15:47] at FsUnit.shouldEqual[a](a x, a y)
```

ncave:
@dsyme Yes, it shouldn't be failing but it is, look at the messages it outputs when it fails. Perhaps as you said the `assumeDotNetFramework `flag needs to be set.

ncave:
@dsyme Just adding the assumeDotNetFramework flag is not enough, I added back the reference from the original PR and now the test is passing again.

src/fsharp/CompileOps.fs
```diff
@@ -1700,6 +1700,7 @@ let DefaultReferencesForScriptsAndOutOfProjectSources(assumeDotNetFramework) =
           yield "System.Numerics" 
       else
           yield Path.Combine(Path.GetDirectoryName(typeof<System.Object>.Assembly.Location),"mscorlib.dll"); // mscorlib
+          yield typeof<System.Object>.Assembly.Location; // System.Private.CoreLib
```
dsyme: We need to get to the bottom of why this is needed. AFAIK it should never be necessary to reference System.Private.CoreLib or any implementation assemblies which in turn refer to this.

I'll admit I find the facade/implementation assembly split bewildering in practice and the details always slip away.  Do you know what's causing the need for this reference?  Thanks

ncave: I don't, but as already mentioned in the original PR, there is a very unequivocal error message saying it's needed:
```
The type 'Object' is required here and is unavailable. You must add a reference to assembly System.Private.CoreLib
A reference to the type 'System.Object' in assembly 'mscorlib' was found, but the type could not be found in that assembly
```
Which kind of points to some issue in resolving forwarded types. Perhaps somebody from the dotnet/core team can shed a light?

ncave: I'm really hoping this issue will go away with .netstandard 2.0, we'll see.

dsyme: > The type 'Object' is required here and is unavailable. You must add a reference to assembly System.Private.CoreLib

I believe this indicates that the compilation is referencing .NET Core private implementation assemblies, which is now considered invalid

In .NET Framework you used to be able to just reference any assemblies under the framework implementation directory ``c:\windows\...`` . But my understanding is that in the .NET Core world they want all compilations only use facade/reference/... assemblies.  I believe this includes dynamic compilations, which is a bit confusing for me since mentally I always think of .NET Framework dynamic compilations (e.g. ``fsi.exe``) as being against the latest installed .NET as found in the implementation directory.  

I'm not at all clear on where/how .NET Core dynamic code generation for tools like ``fsi.exe`` is expected to find the appropriate reference assemblies, it almost seems to me that they expect a tool like ``fsi.exe`` to download all the nuget packages etc. for a particular framework facade and use those, or find cached ones on the machine.

Because of this, when I last looked at ``fsi.exe`` on .NET Core, I ended up trying to [teach the compiler to respect System.Private.CoreLib as a primary assembly reference for System.Object etc](https://github.com/Microsoft/visualfsharp/pull/1792#discussion_r88955380).  But @KevinRansom told me not to, that this is wrong, and compilation should never reference this assembly.  I'm still confused about what they should reference, and I can really see that in practice people doing dynamic compilation are going to reference the implementation assemblies all the time, just because it's dead simple and you can be guaranteed that they actually exist.  

Another way to put this is that I'm simply not sure what the correct values of ``DefaultReferencesForScriptsAndOutOfProjectSources`` should be for .NET Core, i.e. what the default references for scripts should be.  It feels like they should be the locations of a bunch of .NET Standard 2.0 reference assemblies, but I don't think we can assume they are installed.

Anyway, to help us make progress, let's characterize the specific scenario carefully. I assume it is this:

- You are running the .NET Core version of a tool called Fable.EXE (or whatever its name) 

- Your Fable.EXE uses the .NET Core version of FCS.dll to analyze an F#/Fable script or project, extract its Symbols and Exprs, and generate (print) Javascript code. Your Fable.EXE  doesn't actually do any .NET codegen, but rather JS codegen.  Your Fable.EXE may accept references to .NET and F# assemblies as part of its input specification.

- Your Fable.EXE prefers not to require any additional facade or reference assembly packages.  For the moment it's happy to rely on those returned by evaluating these fragments on .NET Core, taken from ``DefaultReferencesForScriptsAndOutOfProjectSources``:

```
yield Path.Combine(Path.GetDirectoryName(typeof<System.Object>.Assembly.Location),"mscorlib.dll"); // mscorlib
         yield typeof<System.Object>.Assembly.Location; // System.Private.CoreLib"
...
```

- We know these assembly references are "wrong" (they are private implementation assemblies) but they appear to "work" because no actual .NET code generation is done, and the F# compilation logic doesn't really care

- We know that it may be better to reference .NET Standard 2.0 reference assemblies or Fable-specific reference assemblies but that adds complexity to the tool and we don't know where to find these

Is that correct?  BTW have you tried adding the above reference explicitly in ``otherArgs`` of ``GetProjectOptionsFromScript``?

thanks
don

ncave:
@dsyme It's not really about Fable, it affects any client that uses checker.GetProjectOptionsFromScript on .NET Core. But yes, adding the reference directly to the script works too.

As you suggested, I have removed the System.Private.CoreLib from the list of default references and added it only at the test script so the test can pass.

tests/service/ProjectAnalysisTests.fs
```diff
@@ -4505,10 +4505,16 @@ module Project35b =
     open System.IO
 
     let fileName1 = Path.ChangeExtension(Path.GetTempFileName(), ".fsx")
+#if DOTNETCORE
+    let fileSource1 = """
+#r "System.Private.CoreLib.dll"
```
KevinRansom: So ... this is fine for a test but not as a long term solution.  We will have a decent solution for this eventually.

A preferable approach for tests would be to use:
https://github.com/Microsoft/visualfsharp/blob/master/tests/scripts/fsci.fsx
with a project.json file identifying dependencies.

Kevin

ncave: @KevinRansom For checker.GetProjectOptionsFromScript, is there a duplication of responsibility between `assumeDotNetFramework=false` and `otherFlags = [| "--targetprofile:netcore" |]` ?

dsyme:
@ncave Not sure why we're got this - am retesting
```
[00:11:02] log  : Restoring packages for C:\projects\fsharp-compiler-service\src\fsharp\FSharp.Compiler.Service.MSBuild.v12\project.json...
[00:11:04] error: Unable to load the service index for source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json.
[00:11:04] error:   Response status code does not indicate success: 503 (Service Unavailable).
[00:11:04] info :   GET https://api.nuget.org/v3-flatcontainer/system.diagnostics.tracesource/index.json
[00:11:04] Running build failed.
[00:11:04] Error:
[00:11:04] System.Exception: Command failed with exit code 1
```

ncave:
Closing as perhaps obsolete.

PR #659, opened 2016-10-22
Fix ProjectCracker, optimize Filename.checkPathForIllegalChars

vasily-kirichenko:
This function popped up in a profiler as one of hot paths. With this PR it disappears. 

`Path.GetInvalidPathChars()` returns 36 chars and checking, say, 200 chars path, means 7200 array accesses and char comparisons. It seems this function is called very often.


dsyme:
CI is failing because on .NET Core 1.0 we have
```
error FS0039: The field, constructor or member 'InvariantCultureIgnoreCase' is not defined
```

vasily-kirichenko:
@dsyme Thanks!

dsyme:
Hmmm On both Travis and AppVeyor: 
```
Errors and Failures:
1) Test Failure : FSharp.Compiler.Service.Tests.ExprTests.Check use of type provider that provides calls to F# code
     Expected: 0
Actual: 3
  Expected: 0
  But was:  3

at FsUnit.shouldEqual[a] (a x, a y) <0x40e11900 + 0x0018b> in <filename unknown>:0
at FSharp.Compiler.Service.Tests.ExprTests.Check use of type provider that provides calls to F# code () <0x40324800 + 0x001c8> in <filename unknown>:0

```

vasily-kirichenko:
Yeah, same failure on my machine...

vasily-kirichenko:
So, `ProjectFileNames` and `OtherOptions` are in a mess. Before this PR `ProjectFileNames` was always empty and files names were in `OtherOptions`. And this worked somehow. I thought the proper way was to _copy_ files names to `ProjectFileNames` from `OtherOptions`. But it messed things up since some code combines files names from those two lists. So, in the last commit I removed file names from `OtherOptions`, but it seems some other code uses `OtherOptions` and ignores `ProjectFileNames` :(

vasily-kirichenko:
OK, I fixed the tests according to this behavior:

* source files go to `ProjectFileNames` _only_

dsyme:
@vasily-kirichenko We should really merge the perf fix into Microsoft\visuafsharp too

dsyme:
@vasily-kirichenko  Thanks for the contribution!

vasily-kirichenko:
@dsyme I ported the perf fix to visualfsharp repo and tested it on FCS project. Absolutely no improvements: 43 seconds on F# 4.0/4.1(master)/this perf fix. So I don't think it's worth to make this change in VFT.

dsyme:
@vasily-kirichenko It might be that the fix applies primarily in the IDE tools - endless rechecking of fie names while preparing project options or doing incremental building.  So perhaps the fix would apply specifically to the Visual F# IDE Tools.  Anyway, it's ok - we can back port at some point, or perhaps we eventually work out how to merge repos.

vasily-kirichenko:
Ah, I see. It's very easy to port it, so I'll make a PR in a minute or two.

vasily-kirichenko:
https://github.com/Microsoft/visualfsharp/pull/1662

PR #662, opened 2016-10-27
ServiceNavigation: fixed wrong range for member val

vasily-kirichenko:
ServiceNavigation: fixed wrong range and duplicates for properties


vasily-kirichenko:
It's fixed Ionide CodeLens feature:

image

dsyme:
@vasily-kirichenko Should really have had a test as part of the PR I suppose....  But merged in any case

Krzysztof-Cieslak:
@dsyme, could we get NuGet release with latest changes, including this one? :-)

Issue #605, opened 2016-07-09
Discussion for: When can we have an alpha release of FCS for dot net core?


7sharp9:
When and how can we get this?

Building and referencing locally is time consuming too.

dsyme:
@7sharp9 We're waiting on a PR from @enricosada to get .NET Core AppVeyor build green and merge the nuget packages

ncave:
@enricosada Is there other pending work after [#616](https://github.com/fsharp/FSharp.Compiler.Service/pull/616) was merged? Does the FCS/ProjectCracker/Tool version need to be incremented before it's released?
Not sure how FCS version jumped from 3.0 to 6.0.1 (already released on NuGet on July 22).

dsyme:
@ncave We're using SemVer now, or attempting to. Each major integration from Microsoft\visualfsharp is likely to cause a bump in the major version number, particularly because changes to the "AST" data structure might cause binary incompatibilites.  So it's safest to just bump the major version number quite often.  I wouldn't be surprised to be at 50.0.0.0 or 100.0.0.0 fairly quickly...

enricosada:
@ncave nothing i know about, but after fable work i think you know better than anyone else :D

ncave:
@enricosada I'm sure there are upstream changes that need to make their way from ms/visualfsharp, but perhaps we can get another alpha release before that if somebody increments the version numbers (not sure who makes the decision on releases, perhaps the maintainers).

dsyme:
Yes - @rneatherway, @forki and myself can make releases.  

I believe most recent changes from Microsoft\visualfsharp have been integrated in 6.0.2

ncave:
@dsyme There may be some upstream commits that are not in FCS yet, like these (and maybe more):

https://github.com/fsharp/fsharp/commit/c5445a48d72704313e90dbba6dd9832355aafb56
https://github.com/fsharp/fsharp/commit/be8b1e238aa6db9443bdd038de6510444de473ab

7sharp9:
Any news on this?

dsyme:
The latest changes from fsharp/fsharp are merged (https://github.com/fsharp/FSharp.Compiler.Service/pull/635/). The nuget package hasn't yet been updated.

Would be great to have a finalized release.  What's needed to do that now? thanks

7sharp9:
fslex,fsyacc dnc versions so it builds locally?

dsyme:
[...]

PR #680, opened 2016-12-10
enable netcore build with #EXTENSIONTYPNG

cloudRoutine:
addressing #605


cloudRoutine:
@dsyme are there any changes that I need to make?

PR #682, opened 2017-01-04
Clear sources in NuGet.Config

forki:
This is important so that we use proper sources.

But after that I still get:

image

@cloudRoutine  I think it should install it's own dotnet core like in https://github.com/fsprojects/Paket/blob/core3/build.fsx#L131-L184


dsyme:
Is this ready to merge?

forki:
Yes it is. Just takes me to the next issue...

dsyme:
thanks :)

Issue #676, opened 2016-12-04
Question: IsImplicitConstructor in Entity methods from signature files


alfonsogarciacaro:
Related to https://github.com/fable-compiler/Fable/issues/571

Fable relies on the property `IsImplicitConstructor` to translate F# code to JS, as ES2015 classes only accept one unique constructor and the others must be compiled as mangled static methods. However when checking a method from an entity declared in a signature file (.fsi) `IsImplicitConstructor` always evaluates to false so the constructor call gets compiled wrongly (as Fable thinks it's a static method).

Would it be possible to have `IsImplicitConstructor` evaluate the same way no matter if it's coming from a signature or an implementation file? Or at least, how can the implementation Entity be accessed from a signature Entity reference? (I haven't found an easy way to do this.)

PR #688, opened 2017-01-21
Fable support

ncave:
- Mostly #ifdefs added, but not too many (IMO)
- Uses a custom build step for code generation: `fcs\build CodeGen.Fable`
- **Parsing multiple files works**
- **Untyped AST parsing works**
- **Typed AST parsing also works**
- **Compiling Fable with Fable works**
- **[Fable2 Demo REPL](http://fable.io/repl2) page works** (Fable1 REPL [here](http://fable.io/repl))
- Next Step: To be determined by community, open for suggestions.

See also: [Fable JS](https://github.com/fable-compiler/Fable/pull/657)


dsyme:
@ncave Extraordinary.   I'll take a close look

Is the demo page hosted somewhere? How big is the overall JS for the bootstrapped compiler?

Random future thought: Do you think there's any chance we could eventually produce .NET binaries with the JS compiler (and send them to .NET clients?)  I've no idea when/where that would be useful, it's just a thought.

Also, I wonder if this might eventually make it into the Microsoft Visual F# repo and the standard F# command line compiler.  That would be quite interesting :)  Or is it always best to package it via Fable?

Export.FCS.sln
```diff
@@ -0,0 +1,27 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
```
dsyme: It's probably best if all files that have something to do with Fable have Fable in the name where possible?

build.fsx
```diff
@@ -272,14 +272,27 @@ Target "CodeGen.NetCore" (fun _ ->
     let fsLex fsl out = runInDir (toolDir + "fslex.exe") "%s --unicode %s -o %s" fsl lexArgs out
     let fsYacc fsy out m o = runInDir (toolDir + "fsyacc.exe") "%s %s %s %s %s -o %s" fsy lexArgs yaccArgs m o out
 
+#if FABLE_COMPILER
```
dsyme: Just to mention that in the visualfsahrp (and soon fsharp, and soon this) repo FsSrGen has been replaced by a script fssrgen.fsx

build.fsx
```diff
@@ -272,14 +272,27 @@ Target "CodeGen.NetCore" (fun _ ->
...
     fsLex "../../absil/illex.fsl" "illex.fs"
     fsYacc "../../absil/ilpars.fsy" "ilpars.fs" module1 open1
     fsYacc "../pars.fsy" "pars.fs" module2 open2
     fsYacc "../pppars.fsy" "pppars.fs" module3 open3
+
+#if FABLE_COMPILER
+    // comments the #line directive as it is not supported by Fable
+    ["lex.fs"; "pplex.fs"; "illex.fs"; "ilpars.fs"; "pars.fs"; "pppars.fs"]
+    |> Seq.map (fun fileName -> IO.Path.Combine (workDir, fileName))
+    |> RegexReplaceInFilesWithEncoding @"# (?=\d)" "//# " Text.Encoding.UTF8
```
dsyme: Any specific reason this doesn't work in fable?

dsyme: Any specific reason this doesn't work in fable?

src/absil/il.fs
```diff
@@ -8,14 +8,18 @@ module (*internal*) Microsoft.FSharp.Compiler.AbstractIL.IL
 
 
 open Internal.Utilities
+#if FABLE_COMPILER
```
dsyme: You may as well make this unconditional e.g.

    open Microsoft.FSharp.Collections // needed for Fable
    open Microsoft.FSharp.Core // needed for Fable

src/absil/il.fs
```diff
@@ -8,14 +8,18 @@ module (*internal*) Microsoft.FSharp.Compiler.AbstractIL.IL
...
+open Microsoft.FSharp.Core
+#endif
 open Microsoft.FSharp.Compiler.AbstractIL
 open Microsoft.FSharp.Compiler.AbstractIL.Diagnostics
 open Microsoft.FSharp.Compiler.AbstractIL.Internal
 open Microsoft.FSharp.Compiler.AbstractIL.Internal.Library
 open System.Collections
 open System.Collections.Generic
 open System.Collections.Concurrent
-
```
dsyme: Please remove whitespace cahnges

src/absil/il.fs
```diff
@@ -77,21 +81,21 @@ let rec splitNamespaceAux (nm:string) =
 
 /// Global State. All namespace splits ever seen
 // ++GLOBAL MUTABLE STATE
+#if FABLE_COMPILER
```
dsyme: Why not conditionally define 

    type ConcurrentDictionary<'K,'V> = Dictionary<'K'V>

?

src/absil/il.fs
```diff
@@ -568,12 +571,14 @@ type ILTypeRef =
         | _ -> ILType.Boxed tspec
 
     override x.GetHashCode() = x.hashCode
-    override x.Equals(yobj) = 
-         let y = (yobj :?> ILTypeRef) 
-         (x.ApproxId = y.ApproxId) && 
-         (x.Scope = y.Scope) && 
-         (x.Name = y.Name) && 
-         (x.Enclosing = y.Enclosing)
+    override x.Equals(yobj:obj) =
```
dsyme: Why is this change needed?  When would the cast fail?

src/absil/il.fs
```diff
@@ -1365,10 +1370,10 @@ type ILMethodDef =
     member x.IsNonVirtualInstance = match x.mdKind with | MethodKind.NonVirtual -> true | _ -> false
     member x.IsVirtual            = match x.mdKind with | MethodKind.Virtual _  -> true | _ -> false
 
-    member x.IsFinal                = match x.mdKind with | MethodKind.Virtual v -> v.IsFinal    | _ -> invalidOp "not virtual"
-    member x.IsNewSlot              = match x.mdKind with | MethodKind.Virtual v -> v.IsNewSlot  | _ -> invalidOp "not virtual"
-    member x.IsCheckAccessOnOverride= match x.mdKind with | MethodKind.Virtual v -> v.IsCheckAccessOnOverride   | _ -> invalidOp "not virtual"
-    member x.IsAbstract             = match x.mdKind with | MethodKind.Virtual v -> v.IsAbstract | _ -> invalidOp "not virtual"
+    member x.IsFinal                = match x.mdKind with | MethodKind.Virtual v -> v.IsFinal    | _ -> false //invalidOp "not virtual"
```
dsyme: Why are these changes needed?  I see they are pretty harmless changes though

src/absil/il.fs
```diff
@@ -3661,7 +3666,11 @@ type ILTypeSigParser(tstring : string) =
                 // fetch the arity
                 let arity = 
                     while (int(here()) >= (int('0'))) && (int(here()) <= ((int('9')))) && (int(peek()) >= (int('0'))) && (int(peek()) <= ((int('9')))) do step()
+#if FABLE_COMPILER
```
dsyme: Can we just always use the Fable version here?

src/absil/il.fs
```diff
@@ -1552,7 +1557,7 @@ and [<Sealed>] ILTypeDefs(f : unit -> (string list * string * ILAttributes * Laz
     let mutable array = InlineDelayInit<_>(f)
     let mutable dict = InlineDelayInit<_>(fun () -> 
             let arr = array.Value
-            let t = Dictionary<_,_>(HashIdentity.Structural)
+            let t = Dictionary<_,_>(3, HashIdentity.Structural)
```
dsyme: What's up here - doesn't Fable have a default size constructor?

dsyme:
Made a few initial comments.  I suppose the big challenge is going to be keeping this compiling and working as changes flow in from the Visual F# TOols

Export.FCS.sln
ncave: Sure. It's not really related to the PR, it's just a quick hack to export metadata using FCS, so it should be moved out.

build.fsx
ncave: I know, it's there just until the new tooling get released and FCS is migrated to use it.

build.fsx
ncave: Yes, see [this](https://github.com/fable-compiler/Fable/issues/400).

src/absil/il.fs
```diff
@@ -395,13 +400,12 @@ type ILAssemblyRef(data)  =
...
         ILAssemblyRef.Create(aname.Name,None,publicKey,retargetable,version,locale)
- 
-
+ #endif
 
     member aref.QualifiedName = 
         let b = new System.Text.StringBuilder(100)
         let add (s:string) = (b.Append(s) |> ignore)
-        let addC (s:char) = (b.Append(s) |> ignore)
+        let addC (s:char) = (b.Append(string s) |> ignore)
```
ncave: Sure

src/absil/il.fs
ncave: Maybe, but that's not the default value for this contructor. In any case, this will probably go away, as it's just a work around a deficiency in Fable (can't do comparers), when it's resolved I'll take it out.

ncave: Main reason is Fable (and Javascript maps/sets) have no proper support for equality comparers. Technically map/set trees with IComparer can be used as a substitute, but probably the performance will not be too good (not sure how important collection performance is here though). Very few places in FCS actually use a real equality comparer though. Technically this should go away as it's just a deficiency in Fable.

src/absil/il.fs
ncave: Sure, I just tried to be conservative with changes to lower the impact.

ncave:
@dsyme @smoothdeveloper Thank you for your comments!
Some of the unexplicable ifdefs are just to get around deficiencies in the compiler, while some others are actually a problem in FCS (see #676 for example, which is forcing the addition of dummy constructors all over the place).

@dsyme
1) Demo page is not hosted anywhere yet but you can host it locally with Node after it's compiled with Fable. The bundled JS size is around 8 MB (unminified) so it should get smaller when minified. Compilation time for scripts is not as bad as I thought it would be, just a few seconds, but after all the whole thing is completely unoptimized, so it should get better when the hot path is optimized.

2) Sure, I see no technical reason to not be able to produce .NET binaries with the JS compiler. Not sure if that's applicable in a browser setting, but perhaps in Node setting. A bit more of FCS will need to be compiled, but shouldn't be a problem if there is interest.

3) Once the compiler is compiled to JS it can be bundled / distributed anywhere, not just in a browser. For a full-featured replacement as a tool it will need the rest of FCS to be compiled as well (file system, build system, project support, etc.).

build.fsx
ncave: Fable uses the line directive already to do source mapping, so we can't. The simplest way is to just comment it out.

src/absil/il.fs
ncave: Will do.

src/absil/il.fs
ncave: No reason, just looked cleaner this way. This too can go away after Dictionary constructors can be handled properly by Fable.

src/absil/il.fs
ncave: I don't recall the actual reason, probably unsupported casting by Fable.

src/absil/il.fs
ncave: I'll revisit that.

dsyme:
@ncave Looks like this is green :) Do you have more work you'd like to do on this, or should we work towards a final review?

ncave:
@dsyme The PR can probably use some work to clean it up a bit, but if I recall correctly, you mentioned that there are some large upstream changes incoming, so it may be a good idea to wait until those are merged in and adapt the PR to them before moving forward.

.vscode/launch.json
```diff
@@ -0,0 +1,21 @@
+{
```
dsyme: Should this be included in the PR?

src/absil/illib.fs
```diff
@@ -20,9 +25,10 @@ let (>>>&) (x:int32) (n:int32) = int32 (uint32 x >>> n)
 let notlazy v = Lazy<_>.CreateFromValue v
 
 let inline isNonNull x = not (isNull x)
-let inline nonNull msg x = if isNull x then failwith ("null: " ^ msg) else x
+let inline nonNull msg x = if isNull x then failwith ("null: " + msg) else x
 let (===) x y = LanguagePrimitives.PhysicalEquality x y
 
+#if !FABLE_COMPILER
```
dsyme: It may be better to use a conditional based on feature like ``#if FX_NO_GETCURRENTPROCESS`` or the like

src/absil/illib.fs
```diff
@@ -36,13 +42,19 @@ let reportTime =
...
+#endif
 
 //-------------------------------------------------------------------------
 // Library: projections
 //------------------------------------------------------------------------
 
-[<Struct>]
 /// An efficient lazy for inline storage in a class type. Results in fewer thunks.
+#if FABLE_COMPILER
+type InlineDelayInit<'T when 'T : not struct>(f: unit -> 'T) =
```
dsyme: When you use ``#if FABLE_COMPILER`` please add a comment indicating what Fable doesn't support, or use a feature define like ``#if FSLANG_NO_MUTABLE_STRUCTS``.  The latter helps document that the F# language as supported by Fable has some specific semantic differences to F# as supported by .NET.

src/absil/illib.fs
```diff
@@ -398,10 +413,10 @@ module String =
         if r = -1 then indexNotFound() else r
 
     let contains (s:string) (c:char) = 
-        s.IndexOf(c,0,String.length s) <> -1
+        s.IndexOf(c) <> -1
 
     let order = LanguagePrimitives.FastGenericComparer<string>
-
```
dsyme: Please remove whitespace changes where possible

src/absil/illib.fs
```diff
@@ -808,10 +831,17 @@ type LayeredMap<'Key,'Value  when 'Key : comparison> = Map<'Key,'Value>
 type Map<'Key,'Value when 'Key : comparison> with
     static member Empty : Map<'Key,'Value> = Map.empty
 
+#if FABLE_COMPILER
```
dsyme: Add explanatory comment for the ``#if`` please

src/absil/ilread.fs
```diff
@@ -1029,377 +1032,373 @@ let count c =
 #endif
         
 
-let seekReadUInt16Adv ctxt (addr: byref<int>) =
```
dsyme: These changes are a problem - the use of ``byref`` was introduced to eliminate a of tuple allocations.  Using a struct tuple or struct return would be acceptable I think (though I'd kind of prefer if we could freely use byref args in the comppiler codebase, where perf can be critical....)

src/absil/ilwrite.fs
```diff
@@ -2546,40 +2545,40 @@ let GenMethodDefAsRow cenv env midx (md: ILMethodDef) =
     if md.IsEntryPoint then 
         if cenv.entrypoint <> None then failwith "duplicate entrypoint"
         else cenv.entrypoint <- Some (true, midx)
-    let codeAddr =
```
dsyme: Please remove whitespace changes

src/fsharp/CompileOps.fs
```diff
@@ -3,25 +3,29 @@
...
+#endif
 open Internal.Utilities.Text
 open Internal.Utilities.Collections
 open Internal.Utilities.Filename
 
-open Microsoft.FSharp.Compiler.AbstractIL 
-open Microsoft.FSharp.Compiler.AbstractIL.IL 
-open Microsoft.FSharp.Compiler.AbstractIL.Internal 
-open Microsoft.FSharp.Compiler.AbstractIL.Internal.Library 
+open Microsoft.FSharp.Compiler.AbstractIL
```
dsyme: Please remove whitespace changes

src/fsharp/Fable.FCS/app.fs
```diff
@@ -0,0 +1,69 @@
+module App
```
dsyme: How about changing ``Fable.FCS`` to ``fcsj``?  It is after all an F# compiler for Javascript.  Or ``fcs-js`` or ``fcs-fable``?  Just whatever is most idiomatic

src/fsharp/Fable.FCS/fsstrings.fs
```diff
@@ -0,0 +1,968 @@
+module internal SR.Resources
```
dsyme: Why does this need to be checked in rather than generated?

src/fsharp/NicePrint.fs
```diff
@@ -103,7 +103,7 @@ module internal PrintUtilities =
         let tcref = attrib.TyconRef
         squareAngleL (layoutTyconRefImpl true denv tcref)
 
-module private PrintIL =
```
dsyme: Why removing "private"??

src/fsharp/Fable.FCS/fsstrings.fs
ncave: Because It's not generated right now, it's just a resource. Perhaps it was generated before.

src/fsharp/NicePrint.fs
ncave: Unfortunately private modules are not yet properly supported in Fable. I'll open an issue to discuss it.

src/fsharp/Fable.FCS/app.fs
ncave: Will do.

src/absil/ilread.fs
ncave: Support needs to be added in Fable for it. I'll open an issue.

.vscode/launch.json
ncave: I'll take it out. It was just handy to have a debugging environment out of the box, but it does not belong in the PR.

src/absil/ilread.fs
dsyme: @ncave I suppose we don't mind a major `#if` in the interim, though it's unfortunate

ncave: @dsyme As JavaScript lang does not support byref parameters unless they're wrapped in objects, there will be no performance advantage in JS. I've opened a Fable feature request for the byref support, but perhaps we can use struct tuples here, unless you prefer #ifdefs to retain the byref for dotnet?

src/fsharp/NicePrint.fs
ncave: @dsyme See #690.

src/absil/ilread.fs
dsyme: @ncave I think use ``#if`` for now.  The code is performance sensitive for startup of the F# compiler and it's best not to disturb it. thanks!

dsyme: (If you write the Fable version using struct-tuples, and we can get it into a PR for visualfsharp, and we can performance test it, then we can switch to that version)

src/absil/il.fs
ncave: It may be a good idea to retain the conditional as there is some type overwriting that is only needed when compiled with Fable.

src/absil/illib.fs
ncave: Looks like `FX_NO_GETCURRENTPROCESS` is not used anywhere else. I've tried to keep the conditionals to a minimum and not introduce new ones unnecessarily. IMO, there are quite a few BCL APIs not supported by Fable, it would become crowded very fast if we introduce a different conditional for each one that is not supported. But if you feel strongly about it, I'll change it.

src/absil/il.fs
ncave: It may be a good idea to retain the conditional as there is some type overwriting that's only needed when compiled with Fable.

src/absil/ilread.fs
ncave: @dsyme This is done (giant #if around the whole thing).

src/absil/illib.fs
dsyme: Ideally I'd like the F# compiler codebase to not mention Fable at all , but rather be ``#if`'d on a set of characteristics that Fable happens to have.  I'm expecting that there will other F#-to-XYZ attempts in the future, e.g. to the JVM, or another F#-to-Javascript compiler (e.g. WebSharper), or F#-to-native.

The more we can use ``FX_NO_XYZ...`` or ``FSLANG_NO_BYREFS`` the more the compiler code will be "neutral" with respect to these different efforts.

src/absil/ilread.fs
```diff
@@ -1028,6 +1036,378 @@ let count c =
     ()
 #endif
         
+#if FABLE_COMPILER // no byref parameters
```
dsyme: For example this one would be nicer as ``FSLANG_NO_BYREFS``, thanks

src/fsharp/MethodCalls.fs
```diff
@@ -68,7 +68,7 @@ type CalledArg =
       NameOpt: Ident option
       CalledArgumentType : TType }
 
-let CalledArg(pos,isParamArray,optArgInfo,callerInfoInfo,isOutArg,nameOpt,reflArgInfo,calledArgTy) =
+let GetCalledArg(pos,isParamArray,optArgInfo,callerInfoInfo,isOutArg,nameOpt,reflArgInfo,calledArgTy) =
```
dsyme: Is this because Fable doesn't let you have a type and a function with the same name in the same module? thanks

src/fsharp/NameResolution.fs
```diff
@@ -1352,6 +1360,9 @@ type TcResultsSinkImpl(g, ?source: string) =
     let capturedMethodGroupResolutions = ResizeArray<_>()
     let allowedRange (m:range) = not m.IsSynthetic       
 
+#if FABLE_COMPILER
```
dsyme: What are these for? thanks

src/fsharp/NicePrint.fs
dsyme: Thanks

src/fsharp/NicePrint.fs
```diff
@@ -1736,7 +1764,7 @@ module private TastDefinitionPrinting =
 
 //--------------------------------------------------------------------------
 
-module private InferredSigPrinting =
```
dsyme: What if these are made internal instead of private?  Does Fable support internal modules? thanks

src/fsharp/QuotationTranslator.fs
```diff
@@ -22,8 +22,11 @@ open System.Collections.Generic
 
 module QP = Microsoft.FSharp.Compiler.QuotationPickler
 
-
+#if FABLE_COMPILER
```
dsyme: Probably best to put the ``#if`` over ``condition`` rather than at callsites

src/fsharp/infos.fs
```diff
@@ -1594,7 +1594,6 @@ type ILFieldInfo =
         | ILFieldInfo(_, x1), ILFieldInfo(_, x2) -> (x1 === x2)
 #if EXTENSIONTYPING
         | ProvidedField(_,fi1,_), ProvidedField(_,fi2,_)-> ProvidedFieldInfo.TaintedEquals (fi1, fi2) 
-        | _ -> false
```
dsyme: Why is this removed?

src/utils/ResizeArray.fs
```diff
@@ -69,17 +69,29 @@ module internal ResizeArray =
         res
 
     let mapi f (arr: ResizeArray<_>) =
+#if !FABLE_COMPILER
```
dsyme: You may like to check if any of these are even used in the compiler

src/fsharp/fcs-fable/adapters.fs
```diff
@@ -0,0 +1,269 @@
+namespace Internal.Utilities
```
dsyme: Please put this under ``src/frameworks/fable/adapters.fs``.  We can then put other framework adapters under ``src/frameworks/netcore/...`` etc. Thanks

src/absil/il.fs
dsyme: Please use ``FSLANG_NO_IMPLICITOPEN`` to indicate a dialect of F# that doesn't have implicit opening of these namespaces, thanks

src/absil/il.fs
```diff
@@ -395,13 +400,17 @@ type ILAssemblyRef(data)  =
...
 
         ILAssemblyRef.Create(aname.Name,None,publicKey,retargetable,version,locale)
- 
+ #endif
 
 
     member aref.QualifiedName = 
         let b = new System.Text.StringBuilder(100)
         let add (s:string) = (b.Append(s) |> ignore)
+#if FABLE_COMPILER
```
dsyme: Make this unconditional?

src/absil/il.fs
```diff
@@ -3699,7 +3708,11 @@ type ILTypeSigParser(tstring : string) =
                       yield grabScopeComponent() // culture
                       yield grabScopeComponent() // public key token
                     ] |> String.concat ","
+#if FABLE_COMPILER
```
dsyme: Please use ``FX_NO_ASSEMBLYNAME`` or add an adapter for the AssemblyName type, thanks

src/absil/ildiag.fs
```diff
@@ -6,6 +6,12 @@ module internal Microsoft.FSharp.Compiler.AbstractIL.Diagnostics
 
 open Internal.Utilities
 
+#if FABLE_COMPILER
```
dsyme: What's the specific thing missing here? thanks

src/fsharp/MethodCalls.fs
ncave: Correct.

src/absil/illib.fs
dsyme: (To give context: we've had this before, e.g. for Silverlight we scattered hundreds of ``#if SILVERLIGHT`` across the codebase.  It took years and years to divide them between the various categories - ones that carried across to .NET Core, and ones that were feature-specific etc.   The lesson I learned from that is to always ask people to be specific about why `#if` creep in: if there are just 5-10 uses of an `#if`` it's ok, but there are about 180 uses of ``if FABLE_COMPILER`` - so we should take the time to divide them into the true underlying reasons now)

src/absil/il.fs
ncave: It's not that there is no implicit opening, it's the order of opening after the adapters are loaded which overwrite some core types.

src/absil/illib.fs
ncave: That's fine, if that's your preference I'll try to replace all #if with feature conditionals.

src/fsharp/NameResolution.fs
ncave: @dsyme This is a work-around for #676, unfortunately. There is more of it here and there. I'll change the conditional to be explicit about it.

src/fsharp/infos.fs
ncave: Merging artifact, I'll take it out.

src/utils/ResizeArray.fs
ncave: I'll try to replace them with adapters.

src/fsharp/NicePrint.fs
alfonsogarciacaro: As there're no assemblies in JS, only files. Fable doesn't make a distinction between `internal` and `public`, so using internal modules should work.

alfonsogarciacaro:
> Also, I wonder if this might eventually make it into the Microsoft Visual F# repo and the standard F# command line compiler. That would be quite interesting :) Or is it always best to package it via Fable?

@dsyme I would be very in favour of adding Fable or a Fable-compiled FCS to the F# compiler pipeline and maybe share tests so we can be sure new FCS versions don't break Fable ;)

dsyme:
@ncave The big integrations have flowed into this repo :)   

I'd love to see this PR get over the line - it's really, really important to capture the great work you've done here.  Please let me know if we can help

Thanks!
don

ncave:
@dsyme The PR was rebased to latest.
I haven't converted the FABLE_COMPILER to feature conditionals yet, it will take a while.

dsyme:
@ncave Great, thanks.

alfonsogarciacaro:
@dsyme I'm really happy to present you the result of @ncave's outstanding work, the first ever F# compiler **fully running on the browser** :tada:

http://fable.io/repl.html

It sill needs work to improve performance. If we can make a wish list, mine would ask for an easy way to generate an assembly containing only the metadata for the types supported by Fable. At the moment, I'm using the ones created by @ncave but they still make 1.5 MB that need to be downloaded before anything (and I had to change the extension to `.txt` to allow gzip compression in Github pages). If we could get rid of anything not supported by Fable, I guess we could make the binaries much lighter.

BTW, minified and gzipped the FCS+Fable JS bundle is around 800KB (it seems to have grown almost 200KB with the latest changes in FCS).

dsyme:
>  If we can make a wish list, mine would ask for an easy way to generate an assembly containing only the metadata for the types supported by Fable.

I presume you would need both a new mscorlib.dll (or System.Runtime.dl)  and FSharp.Core.dll. I think the easiest way is just to compile one of each, perhaps as part of Fable itself?

dsyme:
@alfonsogarciacaro Not sure of any reason for the 200KB increase.  Some large chunks of code were added in F# 4.1 and recent language service work, and some more things made public, but the increase feels outside the bounds of what I'd expect

ncave:
@dsyme @alfonsogarciacaro TastOps picked a dependency on EraseClosures and EraseUnions so I had to include those (previously not included), that explains half of it. Also, compiling for different module loaders (umd vs es2015) tends to have large effect on the resulting script size (es2015 being much smaller), the difference can be easily 1 MB or more (uncompressed). Different babel plugins can also play a role there.

dsyme:
@ncave If you like I will work on the FABLE_COMPILER defines and submit a PR to your branch?  Just let me know - I don't want to duplicate work if you're on it :)

ncave:
@dsyme Sure, by all means, that would be nice. I was mostly doing migration so far. Let me know what other areas I can work on.

ncave:
For example, we can migrate FCS to the new tooling/csproj.

dsyme:
> For example, we can migrate FCS to the new tooling/csproj.

That would be great (though I haven't even been keeping up with the status in this area, besides @enricosada's excellent updates)

dsyme:
@ncave We need to have compile-FCS-to-Javascript under test in some form before we can really integrate it into this repo.  Do you think you could add that?  I'm not sure of the best way to do that.  Ideally we would just add a CI pass that uses Fable to compile FCS to Javascript and runs some smoke tests?

enricosada:
Wait before migrate to new fsproj. Going to finalize it in max two days (conversion will be easier for you, plus real packages)

dsyme:
@enricosada That would be great! and a fantastic test case of the new new new  :)

alfonsogarciacaro:
> Also, compiling for different module loaders (umd vs es2015) tends to have large effect on the resulting script size (es2015 being much smaller), the difference can be easily 1 MB or more (uncompressed). Different babel plugins can also play a role there.

@ncave That's because EcmaScript 2015 + bundling enables tree shaking which can remove most of the unused code. About Babel plugins, at the moment I'm compiling FCS to ES2015 mainly because of JS classes. There were problems when calling methods of a parent class if compiling to ES5. The only Babel plugin I'm using is to convert `let/const` to `var` as apparently Google Chrome V8 engine is still unable to optimize the former.

> We need to have compile-FCS-to-Javascript under test in some form before we can really integrate it into this repo. 

@dsyme We could use [Fable tests](https://github.com/fable-compiler/Fable/tree/narumi/src/tests/Main) (though they use a plugin, which are currently not supported by Fable's JS version). Please also note that adding FCS-to-JS compilation plus tests will probably increase the build time by several minutes and that it also means adding Fable sources to this repo. (That's not to say I'm against it, I do agree that the FCS-to-JS compilation should be tested for every FCS release).

> I presume you would need both a new mscorlib.dll (or System.Runtime.dl) and FSharp.Core.dll. I think the easiest way is just to compile one of each, perhaps as part of Fable itself?

@dsyme Yes, Fable would need the metadata for the types in the BCL and FSharp.Core it's compatible with. I guess it's not very important if these are in one or multiple assemblies as long as they're light. At the moment in order to avoid having to copy the full assemblies (you can see the list the online REPL uses [here](https://github.com/fable-compiler/Fable/blob/69e6a57a2d0fe40dacfd3d3c313437eab581ed06/src/dotnet/Fable.Client.Browser/demo/worker.js#L6-L17)), creating the metada-only binaries entails compiling a custom FCS build created by @ncave ([here](https://github.com/ncave/FSharp.Compiler.Service/blob/fable_poc/Export.FCS.sln)) where [these constants are uncommented](https://github.com/ncave/FSharp.Compiler.Service/blob/4f75bd88b5199ae01eef9887caa390fdad9daaae/src/fsharp/FSharp.Compiler.Service/FSharp.Compiler.Service.fsproj#L52-L59).

It'd be nice if the master FCS could expose a method to create these metadata-only binaries, also with a way to filter the types that go into the assembly. Basically Fable only needs the ones in the [compatibility doc](http://fable.io/docs/compatibility.html).

7sharp9:
For the mscorlib etc would we (royal we) rolling a slim one or using a tiny
one from dnc or pcl etc?

dsyme:
> Please also note that adding FCS-to-JS compilation plus tests will probably increase the build time by several minutes 

That's OK, it's "just" CI

> it also means adding Fable sources to this repo

Can't we acquire Fable as an NPM package during CI?

> @dsyme Yes, Fable would need the metadata for the types in the BCL and FSharp.Core 
> it's compatible with. I guess it's not very important if these are in one or multiple assemblies 
> as long as they're light. At the moment in order to avoid having to copy the full assemblies 
> (you can see the list the online REPL uses here), creating the metada-only binaries entails 
> compiling a custom FCS build created by @ncave (here) where these constants are uncommented.

> It'd be nice if the master FCS could expose a method to create these metadata-only binaries, 
> also with a way to filter the types that go into the assembly. Basically Fable only needs the 
> ones in the compatibility doc.

I'm not quite sure of the specific functionality that would go FCS.  It feels like the responsibility to build the FSharp.Core and BCL facades that Fable supports should be in the Fable repo, rather than here?  @ncave What do you think?

dsyme:
> For the mscorlib etc would we (royal we) rolling a slim one or using a tiny one from dnc or pcl etc?

I would expect each Fable release to come with a set of facade assemblies corresponding to the fragment of the BC supported.  I would be surprised if this matched any particular .NET Standard or PCL well enough to be reliable.

ncave:
@dsyme IMO, when netstandard 2.0 comes out, we should be using those binaries, as long as there are no legal issues altering/distributing them.

dsyme:
> IMO, when netstandard 2.0 comes out, we should be using those binaries, as long as there are no legal issues altering/distributing them.

@ncave  I'm sure there will definitely be no legal issues.  But will Fable really support all those types and methods?

ncave:
@dsyme No, but I don't see how they can be easily separated. For .NET Core, they're already broken down in smaller assemblies, so we can pick and choose. And more BCL support can be added in Fable eventually. And they're not *that* big, and they compress well.

dsyme:
> @dsyme No, but I don't see how they can be easily separated. 

I was thinking that Fable would just explicitly write out the necessary facades in C# or F#.

> For .NET Core, they're already broken down in smaller assemblies, so we can pick and choose. 

yes, you're right

> And more BCL support can be added in Fable eventually. And they're not that big, and they compress well.

I was mainly thinking about correctness (only being able to use exactly what's supported) and simplicity (only seeing what's really supported)

However I'm not in the main loop on Fable.  Just let me know what I need to do :)

ncave:
@dsyme IMO the best guarantee for correctness would be to use the same exact officially released binaries that everybody else uses in the dotnet framework (possibly stripped down, without code/resources). This is one of the key features in F# (vs say OCaml), regardless of the target platform (JS or not). We should be expanding support, not contracting it, or at least that should be the goal (IMO).

dsyme:
> @dsyme IMO the best guarantee for correctness would be to use the same exact officially released binaries that everybody else uses in the dotnet framework (possibly stripped down, without code/resources). This is one of the key features in F# (vs say OCaml), regardless of the target platform (JS or not). We should be expanding support, not contracting it, or at least that should be the goal (IMO).

Totally makes sense. The .NET Standard 2.0 binaries (Apart from FSharp.Core) should just be facades (no code/resources) or so I believe

ncave:
@alfonsogarciacaro

> That's because EcmaScript 2015 + bundling enables tree shaking which can remove most of the unused code

I was actually referring to compiled javascript before bundling, i.e. compiling [FCS](https://github.com/ncave/FSharp.Compiler.Service/tree/fable/src/fsharp/fcs-fable) for umd modules is about 1MB bigger than with es2015 modules (uncompressed).

alfonsogarciacaro:
> Can't we acquire Fable as an NPM package during CI?

@dsyme Yeah, this is possible for Fable as an app. But if you want to have an F#-to-JS compiler (not only an F# parser running on JS) you need also [Fable sources](https://github.com/fable-compiler/Fable/blob/69e6a57a2d0fe40dacfd3d3c313437eab581ed06/src/dotnet/Fable.Client.Browser/Fable.Client.Browser.fsproj#L194), but I guess they could also be made available through a package.

> IMO the best guarantee for correctness would be to use the same exact officially released binaries that everybody else uses in the dotnet framework (possibly stripped down, without code/resources). This is one of the key features in F# (vs say OCaml), regardless of the target platform (JS or not). We should be expanding support, not contracting it, or at least that should be the goal (IMO).
(as long as it's modular and you can choose what dependencies to carry).

@ncave I totally agree with you. I was mainly concerned about the size of the assemblies that need to be downloaded (and also having the IDE tell you what's supported or not would be nice) but if the .NET Standard 2.0 are sufficiently small I guess we can wait.

I personally think trying to make Fable another .NET Standard platform is too ambitious, because of the work involved and because many things are not available in a browser environment (file system, threading...) although the .NET Standard team has said it's OK for platforms to throw non-implemented exceptions. But I was thinking the same of compiling FCS with Fable just a month ago so who knows ;)

7sharp9:
@ncave / @alfonsogarciacaro Which parts of FCS are currently functioning here, what are the entry points that are currently being used?   Part of `SourceCodeServices` ?

ncave:
@7sharp9
Single [entry point](https://github.com/ncave/FSharp.Compiler.Service/blob/fable/src/fsharp/fcs-fable/service_shim.fs#L772), parsing & type checking single file.
[Here](https://github.com/fable-compiler/Fable/blob/narumi/src/dotnet/Fable.Client.Browser/Fable.Client.fs#L198-L208) is the client usage for it.

forki:
shouldn't this PR move to VF# now?

dsyme:
@forki Yeah - I've been discussing that with @KevinRansom @cartermp and others offline.  Everyone generally wants that to happen, and to maintain maximal alignment, but there are definitely pros and cons - among other things engineering in VF# is already really complicated and adding cross-compiling to JS with Fable as part of the matrix could be [a bridge too far](https://www.youtube.com/watch?v=0XCmQzfjuK0) today.  We don't even have Mono properly nailed in that repo yet (the mono compilation has some catastrophic bugs, fixed in fsharp/fsharp, but we need to reverse-integrate those fixes), and engineering of .NET Core also needs serious work (e.g. we don't have a Linux/Mac CI for .NET Core in VF#).

For now let's focus on getting this PR into great shape including CI and smoke testing.  Anything we do here will be readily transferrable to VF#.

ncave:
@dsyme @alfonsogarciacaro Rebased to latest (11.01).

ncave:
@alfonsogarciacaro I see you downvoted, is there an issue?

ncave:
Rebased to latest (11.04).

ncave:
Rebased to latest (11.0.8)

ncave:
Rebased to latest (12.0.2)

ncave:
Rebased to latest (12.0.5)

ncave:
Rebased to latest.

enricosada:
thx a lot @ncave to continue to update this. is an amazing work!

dsyme:
Absolutely!

@ncave Would you like to use a branch of this repo for the mainline of this work?

ncave:
@dsyme @enricosada Thank you, it's not hard to maintain as it's basically just a bunch of ifdefs, all credit belongs to the maintainers of FCS and Fable.
- Either way works, fork is fine and branch is fine too, I'll leave it to your preference. I'm not sure what the timeline of the effort to merge FCS back to upstream is, if it's not very far we can keep it in limbo until then. In the mean time, we can try adding a build target and a few tests for it that can be incorporated in the build.
- I keep my ears open for community suggestions if further work is warranted (e.g. full port of incremental builder / project system / IL write / etc. as a tool in a nodejs setting) or should efforts go more towards building a better repl on monaco editor with autocompletion / tooltips.

alfonsogarciacaro:
Hi! Sorry, I've been also busy with conferences and Fable 1.0 issues and couldn't check the REPL recently. The steps I want to make at the moment are:

- Create a build target in Fable repo so it's possible to build the REPL without external dependencies, including the assemblies (atm it's necessary to have several things installed in the local disk). This will allow other contributors help with the REPL.
- From that build target, create an npm package that can be used in a browser or nodejs settings, to allow scenarios like tutorials with editable&executable samples ([reference](https://github.com/fable-compiler/Fable/issues/727#issuecomment-288614194)).
- After that, Monaco & Ionide-web integration :smile:

ncave:
Rebased to latest.

ncave:
Rebased to latest (14.0.2).

ncave:
Rebased to latest (15.0.1).

dsyme:
16.0.1 now in repo too

ncave:
Rebased to latest (16.0.1).

ncave:
Rebased to latest (16.0.2).

forki:
What do we need to get this in?

dsyme:
@forki It is hard.  To integrate into Microsoft/visualfsharp we would need to have it under CI in that repo.  But that means the development in that repo would take a **big** new set of dependencies - on the full Fable compiler - which is as you know still very much under development.

We could integrate it into this repo, but we would be faced with the same problem, and in addition it puts the integration pain on me, when I'm not in a position to fix problems.

I think maintaining a branch makes most sense.  I'd press accept if this gets targets to a "fable" branch and is green under CI.

I'm also happy to just leave this CI open as the active branch - in some ways that's more visible in any case.

cartermp:
I think we're a bit further along in potentially taking this in now than a few months ago, now that we run Fable on the latest .NET Core builds with daily manual validation.  But it's still a very big change.  I'm still in favor of doing that in the long run, but we first need to stabilize our own codebase so that shipping isn't such a hectic effort for us.  Changes are afoot to allow that to happen 😉

dsyme:
Note to self about what makes this so big

* 1486  fcs/fcs-fable/test/package-lock.json
* 1000 lines in FSStrings
* 500 or so in project files etc.

That's about 3/5 of the change.  The other 2000 lines are ``#if`` and quite a large "service shim".  I don't fully understand why the service shim is needed, I suppose it's a simplified version of the current service API without the threading etc.

ncave:
>  I suppose it's a simplified version of the current service API without the threading etc.

Correct, but also loading binaries, simplified setup, etc.

forki:
Lock.json? That should go anyway, right?

forki:
@cartermp so this daily testing  is no longer NDA, right? Can I advertise aggressively?

cartermp:
@forki I'd hold off.  I this month's blog post I plan on having a section saying that.

forki:
Aye

Am 08.10.2017 20:04 schrieb "Phillip Carter" <notifications@github.com>:

ncave:
Rebased to latest (17.0.1).

ncave:
Rebased to latest.

ncave:
Rebased to latest (20.0.1).

ncave:
Rebased to latest (22.0.1).

7sharp9:
@ncave What the status on this so far, if I were to spend some time fitting this into an iOS app would it have many features?

ncave:
@7sharp9 The features are what is listed at the PR head. I've kept it up to date with code merged from upstream, so it's always operational on the latest FCS (even though it says WIP). There are also older version branches in my fork in the unlikely case that you need them.

It would be fairly easy to add more features and even match full FCS functionality if so desired, especially for somebody familiar with the FCS codebase. Please keep in mind that the performance of the resulting JavaScript code is about 5 to 10 times slower than running FCS on dotnet, and (at least for now) below some other competing JavaScript compilers for ML languages (although the hope is performance will increase somewhat with Fable 2.0).

Without knowing what your app functionality requires, here are some links to test projects that showcase the usage:
- A test project for FCS only:
  https://github.com/ncave/FSharp.Compiler.Service/tree/fable/fcs/fcs-fable/test
- A benchmark for Fable + FCS:
  https://github.com/fable-compiler/Fable/tree/master/src/dotnet/Fable.JS/bench
- Some performance discussion:
   https://github.com/fable-compiler/Fable/issues/1205

Let me know if I can help answering any questions about those.

ncave:
Rebased to latest (23.0.1).

ncave:
Rebased to latest (24.0.1).

ncave:
Switched this PR to Fable2 (see other branches for Fable1 support).

ncave:
@dsyme All credit for the new Fable REPL belongs with @MangelMaxime and @alfonsogarciacaro.
Here is the [repo](https://github.com/fable-compiler/repl2) to open enhancement issues (or suggestions) against.

MangelMaxime:
@dsyme Thank you :)

@ncave You are also part of the REPL without your work on the FCS we would not be able to do it :)

@dsyme Yes, it could be possible if we add a server being the REPL to host the generated JavaScript. Currently, the REPL is working only on the browser. Please feel free to open an issue on the repo so we can discuss there.

dsyme:
Thanks all, see https://github.com/fable-compiler/repl2/issues/48 :)

ncave:
@alfonsogarciacaro This has been discussed before. Of course we can, and would be nice to not have to wait until latest (and greatest) gets merged downstream, but it has very little chance (if any) for this PR to ever be merged over there. Not that it has much better chances here either (for understandable reasons), so yeah, perhaps we can just open a new one in a `Microsoft/visualfsharp` fork without submitting a PR. I'll consider it.

ncave:
Rebased to latest (26.0.1).

ncave:
Rebased to latest (26.0.1).

ncave:
Rebased to latest.

ncave:
Rebased to latest (27.0.1).

ncave:
Rebased to latest (28.0.0).

ncave:
Closing this PR in favor of [this one](https://github.com/ncave/fsharp/pull/1) upstream.
It was getting long in the tooth anyway ... 😀

PR #696, opened 2017-02-11
SemanticClassificationType should not be internal

cloudRoutine:
In [service.fsi](https://github.com/fsharp/FSharp.Compiler.Service/blob/master/src/fsharp/vs/service.fsi#L159-L171) `SemanticClassificationType` is marked internal yet
```fsharp
member GetSemanticClassification : unit -> (range * SemanticClassificationType)[]
```
is public and uses it as one of its return types, which makes this member unusable from FSAC and kills the colorization functionality.

### DIFF
--- a/src/fsharp/vs/service.fsi
+++ b/src/fsharp/vs/service.fsi
@@ -157,7 +157,7 @@ type FSharpSymbolUse =
     member RangeAlternate: range
 
 [<RequireQualifiedAccess>]
-type internal SemanticClassificationType =
+type (*internal*) SemanticClassificationType =
     | ReferenceType
     | ValueType
     | UnionCase


vasily-kirichenko:
Yes, it should be fixed. FSAC has never had colorization functionality though.

cloudRoutine:
we don't use it for Ionide, but it has been there. I don't know if the other editor plugins make use of it

[`member __.GetExtraColorizations = checkResults.GetExtraColorizationsAlternate()`](https://github.com/fsharp/FsAutoComplete/blob/master/src/FsAutoComplete.Core/CompilerServiceInterface.fs#L147)

[there are a bunch of integration tests for them too](https://github.com/fsharp/FsAutoComplete/tree/master/test/FsAutoComplete.IntegrationTests/Colorizations)

7sharp9:
I believe they were turned off for performance reasons in VS.

vasily-kirichenko:
The old method returned classification for computation expression names and custom operations usages only. It was used in VFT 2015, but not in VFPT. VFPT has richer colorization code, but I don't like to port it to VFT, VFT uses a bit lower level and 10x simpler approach (which is exposed as this new `GetSemanticClassification` method). I think it should be used by other editors (and FSharp.Formatting) instead of porting VFPT code.

Krzysztof-Cieslak:
@dsyme, can we get NuGet release of this fix as soon as possible? :)

dsyme:
How about grabbing it off the AppVeyor build?

forki:
They want to use it in ionide and fsac. I think that should be a blessed
version then.

Am 13.02.2017 16:59 schrieb "Don Syme" <notifications@github.com>:

forki:
Is this already merged back to Visualfsharp?

vasily-kirichenko:
@forki it's not needed there, but I think no.

forki:
not needed - but we should keep it consistent. otherwise things will conflict and merge errors will happen. and doom.

so: https://github.com/Microsoft/visualfsharp/pull/2440

PR #703, opened 2017-02-20
Integrate visualfsharp and fsharp master

dsyme:


dsyme:
@ncave @forki  Travis on FCS seems to be totally horked because the .NET Core package set downloads are too large, perhaps hitting some networking throttling limitation https://s3.amazonaws.com/archive.travis-ci.org/jobs/203872265/log.txt

Do you know if there's anything we can do to reduce the total download size required for the Ci build?  I presume this is going to be a recurring problem with .NET Core and .NET Standard?

Thanks !

forki:
https://github.com/fable-compiler/fable-suave-scaffold is what I'm using. It seems to work on Travis

ncave:
@dsyme No idea, I know Fable is [installing](https://github.com/fable-compiler/Fable/blob/narumi/build.fsx#L211) the .NET Core sdk on every build (if not there), so that may be one way of doing it.

PR #712, opened 2017-02-28
Trim dependencies

dsyme:
FSSrgen is no longer needed as a package


smoothdeveloper:
@dsyme I see we are using paket here, also seeing `NuGet.Config` plus few `packages.config` in that repository, can you confirm if those are still needed?

dsyme:
@smoothdeveloper  thanks will try removing those

PR #724, opened 2017-03-16
Make incremental builder counter atomic

### LINKED ISSUE #716: Some calls to FCS stops working
I've been trying to investigate https://github.com/ionide/ionide-vscode-fsharp/issues/313. Long story short after using FSAC for a while some requests (autocomplete, tooltips) stops working. As far as I understand issue was there for quite a long time, but only recently I was able to "reproduce" it on my machine.

I've been running FSAC in debug mode to get some additional logging (both FSAC and FCS) and got following results: 

* Normally, when tooltips are working FCS is logging reactor operation `GetToolTipText` and returns result:

* When FSAC "hangs" we still are able to parse file and get errors etc (and all reactor operations are logged by FCS). But when FSAC tries to get tooltips or completions... nothing happens. Nothing is logged by FCS, and operation returns default value `FSharpToolTipText []`.  Following gif shows this behavior

fsac

Sorry for lack of good reproduction, but I don't have it yet. Will try to debug it more later today / tomorrow, but if anyone has any ideas of the top of the head I'd be grateful.

Krzysztof-Cieslak:
Fix #716

### DIFF
--- a/src/fsharp/vs/IncrementalBuild.fs
+++ b/src/fsharp/vs/IncrementalBuild.fs
@@ -1367,7 +1367,7 @@ type IncrementalBuilder(ctokCtor: CompilationThreadToken, frameworkTcImportsCach
     let assertNotDisposed() =
         if disposed then  
             System.Diagnostics.Debug.Assert(false, "IncrementalBuild object has already been disposed!")
-    let mutable referenceCount = 0
+    let referenceCount = ref 0
 
     //----------------------------------------------------
     // START OF BUILD TASK FUNCTIONS 
@@ -1666,17 +1666,17 @@ type IncrementalBuilder(ctokCtor: CompilationThreadToken, frameworkTcImportsCach
 
     member this.IncrementUsageCount() = 
         assertNotDisposed() 
-        referenceCount  <- referenceCount  + 1
+        System.Threading.Interlocked.Increment(referenceCount) |> ignore
         { new System.IDisposable with member x.Dispose() = this.DecrementUsageCount() }
 
     member this.DecrementUsageCount() = 
         assertNotDisposed()
-        referenceCount  <- referenceCount  - 1
-        if referenceCount = 0 then 
+        System.Threading.Interlocked.Decrement(referenceCount) |> ignore
+        if !referenceCount = 0 then 
                 disposed <- true
                 disposeCleanupItem()
 
-    member __.IsAlive = referenceCount > 0
+    member __.IsAlive = !referenceCount > 0
 
     member __.TcConfig = tcConfig
     member __.FileParsed = fileParsed.Publish
@@ -1901,5 +1901,5 @@ type IncrementalBuilder(ctokCtor: CompilationThreadToken, frameworkTcImportsCach
         | Some builder -> builder.IncrementUsageCount() 
         | None -> { new System.IDisposable with member __.Dispose() = () }
 
-    member builder.IsBeingKeptAliveApartFromCacheEntry = (referenceCount >= 2)
+    member builder.IsBeingKeptAliveApartFromCacheEntry = (!referenceCount >= 2)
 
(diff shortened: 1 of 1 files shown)


src/fsharp/vs/IncrementalBuild.fs
```diff
@@ -1666,17 +1666,17 @@ type IncrementalBuilder(ctokCtor: CompilationThreadToken, frameworkTcImportsCach
...
-        referenceCount  <- referenceCount  + 1
+        System.Threading.Interlocked.Increment(referenceCount) |> ignore
         { new System.IDisposable with member x.Dispose() = this.DecrementUsageCount() }
 
     member this.DecrementUsageCount() = 
         assertNotDisposed()
-        referenceCount  <- referenceCount  - 1
-        if referenceCount = 0 then 
+        System.Threading.Interlocked.Decrement(referenceCount) |> ignore
+        if !referenceCount = 0 then
```
0x53A: This may race if someone else modifies it between these calls. I think the return value of ``Interlocked.Decrement`` should be used instead

src/fsharp/vs/IncrementalBuild.fs
```diff
@@ -1901,5 +1901,5 @@ type IncrementalBuilder(ctokCtor: CompilationThreadToken, frameworkTcImportsCach
         | Some builder -> builder.IncrementUsageCount()
```
0x53A: What happens if it is decremented to 0 between getting it, and incrementing the count?
Then it may have been disposed already, but this would set the counter to 1 again.

I think you would need a central routine, which increments the counter first, checks if it was > 0 before, then returns Some (or None if it was at zero).

src/fsharp/vs/IncrementalBuild.fs
```diff
@@ -1367,7 +1367,7 @@ type IncrementalBuilder(ctokCtor: CompilationThreadToken, frameworkTcImportsCach
     let assertNotDisposed() =
         if disposed then  
             System.Diagnostics.Debug.Assert(false, "IncrementalBuild object has already been disposed!")
-    let mutable referenceCount = 0
+    let referenceCount = ref 0
```
0x53A: I think the count should start at 1. Decrementing it to 0 triggers a dispose, and incrementing it from 0 to 1 is illegal, because then the object has already been disposed in another thread..

src/fsharp/vs/IncrementalBuild.fs
dsyme: this is correct

src/fsharp/vs/IncrementalBuild.fs
dsyme: I think no one has access before the first increment.

dsyme: (Nit: This doesn't need to be a ref - you can take the address of the mutable field)

0x53A: Yes, but if you start at 1, then you can later make it illegal to increment from 0 to 1. (Because at 0, the object has already been disposed). This is just a preference.

dsyme: For this PR we're just fixing the race condition, so let's stick with 0 :)

Krzysztof-Cieslak:
Pushed update:
* Reversed backed to mutable value and pass address to Interlocked messages
* Use returned value in `DecrementUsageCount`

0x53A:
I think it can still race at places like this: https://github.com/fsharp/FSharp.Compiler.Service/blob/b95f6fa386405ffba0cae9e3d6d60302dcaf2a2c/src/fsharp/vs/service.fs#L2569

It is first taken from the cache, then the count is incremented. But someone else could have disposed it between these two lines.

dsyme:
@0x53A I see.  Perhaps getOrCreateBuilder and the creation process should return the decrementer

Krzysztof-Cieslak:
I think I'd like to get this merged as soon as possible as it's critical fix for Ionide, and, as far as I can tell, it's "good enough".

We can investigate more fixes later, if it's still problem.

dsyme:
@Krzysztof-Cieslak yes I will merge once green and fix up when releasing

dsyme:
@0x53A thanks for the careful review

Krzysztof-Cieslak:
Travis partial failures ... seems unrelated ;)

rmunn:
@Krzysztof-Cieslak If there are unrelated failures, you can "kick" the build agents by closing and reopening a PR. This will cause all the build agents to rebuild, and you can get rid of a spurious build failure that way.

PR #725, opened 2017-03-16
[Ready for Review] Make incremental builder thread-safe: continuation

0x53A:
This is a continuation of https://github.com/fsharp/FSharp.Compiler.Service/pull/724#issuecomment-287147119

There are three assumption, which i would like to get confirmed @dsyme :

* ``incrementalBuildersCache`` is protected by a Token, so only one thread may access it at a time.
* the last reference is decremented when the builder is pushed out of the cache, so any builder inside the cache is guaranteed to be alive.
* a builder is only pushed out of the cache on the Token-Thread, so there is no chance that the builder is disposed between retrieving it from the cache and incrementing it.

These three assumptions are inside ``getAndIncrementOrCreateBuilder ``.

All other code assumes nothing and may be called concurrently.

------------

The general strategy is that the builder is now incremented inside ``getAndIncrementOrCreateBuilder `` instead of afterwards. For the disposable, which decrements it again, I used the third tuple item, which was the disposable for the cache-eviction before and was always ignored. The cache-eviction disposable is now not returned, but that was kind of pointless anyway.

------------

``TryIncrementUsageCount`` is possibly overly complex - it accounts for a builder in a Schrödinger state (unknown if already disposed or not), but as far as I can see, all used builders are known-alive.
They are retrieved from ``getAndIncrementOrCreateBuilder ``, which is treadsafe anyway, and not used or passed around after disposal.

-------------

Hm ....

```
1) Error : FSharp.Compiler.Service.Tests.PerfTests.Test request for parse and check doesn't check whole project
System.Exception : internal cache error: it should have been impossible for the builder to get disposed at this point!
   at Microsoft.FSharp.Compiler.SourceCodeServices.BackgroundCompiler.getAndIncrementOrCreateBuilder(CompilationThreadToken ctok, FSharpProjectOptions options, CancellationToken ct)
   at <StartupCode$FSharp-Compiler-Service>.$Service.CheckFileInProject@2735-2.Invoke(CompilationThreadToken ctok, CancellationToken _arg10)
```

Either my assumptions are wrong, or it is just too late. Will take a look at this tomorrow again.

### DIFF
--- a/src/fsharp/vs/IncrementalBuild.fs
+++ b/src/fsharp/vs/IncrementalBuild.fs
@@ -27,6 +27,20 @@ open Internal.Utilities
 open Internal.Utilities.Collections
 
 
+module private DisposableHelper =
+    open System.Diagnostics
+
+    let makeOnceDisposable cb =
+        let wasDisposed = ref 0
+        { new IDisposable with
+            member x.Dispose() =
+                if 0 = System.Threading.Interlocked.CompareExchange(wasDisposed, 1, 0) then
+                    cb()
+                else
+                    Debug.Assert(false, "detected double-dispose!")
+        }
+
+
 [<AutoOpen>]
 module internal IncrementalBuild =
 
@@ -1338,7 +1352,7 @@ type IncrementalBuilder(tcGlobals,frameworkTcImports, nonFrameworkAssemblyInputs
     let assertNotDisposed() =
         if disposed then  
             System.Diagnostics.Debug.Assert(false, "IncrementalBuild object has already been disposed!")
-    let mutable referenceCount = 0
+    let mutable referenceCount = 1
 
     //----------------------------------------------------
     // START OF BUILD TASK FUNCTIONS 
@@ -1641,17 +1655,19 @@ type IncrementalBuilder(tcGlobals,frameworkTcImports, nonFrameworkAssemblyInputs
     let MaxTimeStampInDependencies cache (ctok: CompilationThreadToken) (output:INode) = 
         IncrementalBuild.MaxTimeStampInDependencies cache ctok output.Name partialBuild 
 
-    member this.IncrementUsageCount() = 
-        assertNotDisposed() 
-        System.Threading.Interlocked.Increment(&referenceCount) |> ignore
-        { new System.IDisposable with member x.Dispose() = this.DecrementUsageCount() }
+    /// Try to increment the usage count. If it returns None, then the object has been disposed by someone else already and may no longer be used.
+    member this.TryIncrementUsageCount() =
+        ConcurrentIncrement.tryIncrementOrNoneIfZero (&referenceCount)
+        |> Option.map (fun _ -> assertNotDisposed() ; DisposableHelper.makeOnceDisposable this.DecrementUsageCount)
+            
 
     member this.DecrementUsageCount() = 
         assertNotDisposed()
-        let currentValue =  System.Threading.Interlocked.Decrement(&referenceCount)
-        if currentValue = 0 then 
-                disposed <- true
-                disposeCleanupItem()
+        System.Diagnostics.Debug.Assert(referenceCount > 0, "referenceCount <= 0, that means Decrement was called more often than Increment!")            
+        let newValue =  System.Threading.Interlocked.Decrement(&referenceCount)
+        if newValue = 0 then 
+            disposed <- true
+            disposeCleanupItem()
 
     member __.IsAlive = referenceCount > 0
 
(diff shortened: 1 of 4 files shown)


dsyme:
The direction looks right - have you looked at the AppVeyor failure? thanks

0x53A:
NieR: Automata was released yesterday, so no, I haven't taken a deep look at the  failure. =)

I will continue working on this tomorrow evening.

Do you know off the top of your head, if my assumptions are correct with regard to the token and thread safety?

Glory to mankind!

src/fsharp/vs/IncrementalBuild.fs
```diff
@@ -1664,17 +1664,29 @@ type IncrementalBuilder(ctokCtor: CompilationThreadToken, frameworkTcImportsCach
...
+        else
+            let newValue = oldValue + 1
+            let exchanged = System.Threading.Interlocked.CompareExchange(&referenceCount, oldValue, newValue)
+            if exchanged = oldValue then
+                assertNotDisposed()
+                Some { new System.IDisposable with member x.Dispose() = this.DecrementUsageCount() }
+            else if exchanged = 0 then None
+            else
+                // someone else changed 'referenceCount' in the meantime - try again
+                this.TryIncrementUsageCount()
```
dsyme: Why is this so complex again? Why not just this?

```
 member this.IncrementUsageCount() = 
        System.Threading.Interlocked.Increment(&referenceCount) |> ignore
        { new System.IDisposable with member x.Dispose() = this.DecrementUsageCount() }
```
or this

```
 member this.TryIncrementUsageCount() = 
     if System.Threading.Interlocked.Increment(&referenceCount) > 1 then 
        Some { new System.IDisposable with member x.Dispose() = this.DecrementUsageCount() }
     else 
         None
```

0x53A: It is probably over-complex. I was trying to account for a builder in a schrödinger state, where you have a reference to the object, but don't know if it is still alive.

Lets take a builder, which has been disposed, and the ref count is 0.

Two threads enter Increment. The first increments from 0 to 1, and correctly detects that the object is actually dead. BUT the second will then increment it from 1 to 2 and incorrectly revive a dead builder.

The loop is a standard way to deal with this, by reading the initial value, incrementing it, and using compare-exchange to set it. If another thread changed the value between reading it and updating it, the operation is retried.

0x53A:
Nice, ci has been unbroken

src/fsharp/vs/IncrementalBuild.fs
```diff
@@ -1664,17 +1677,56 @@ type IncrementalBuilder(ctokCtor: CompilationThreadToken, frameworkTcImportsCach
...
+        *)
+        // If it goes into the second branch, then someone else updated the value between us reading it and calling CEX, so we need to retry the operation, rereading the value.
+        let initialValue = referenceCount
+        if initialValue = 0 then None
+        else
+            let computedValue = initialValue + 1
+            let exchanged = System.Threading.Interlocked.CompareExchange(&referenceCount, computedValue, initialValue)
+            if exchanged = initialValue then
+                assertNotDisposed()
+                Some <| DisposableHelper.makeOnceDisposable this.DecrementUsageCount
```
fsgit: Please no ``<|``, just use parens in the compiler

forki:
What's the status here?

0x53A:
@forki

Needs a rebase, will do that soon™.


In general, this fix is not as important as I originally thought, after https://github.com/fsharp/FSharp.Compiler.Service/pull/724 there are afaik only two places remaining, with a really unlikely race condition, but it should probably be fixed anyway, the lower the chance of the race, the harder it is to debug an actual failure ...

src/fsharp/vs/IncrementalBuild.fs
```diff
@@ -1641,17 +1655,56 @@ type IncrementalBuilder(tcGlobals,frameworkTcImports, nonFrameworkAssemblyInputs
     let MaxTimeStampInDependencies cache (ctok: CompilationThreadToken) (output:INode) = 
         IncrementalBuild.MaxTimeStampInDependencies cache ctok output.Name partialBuild 
 
-    member this.IncrementUsageCount() = 
-        assertNotDisposed() 
-        System.Threading.Interlocked.Increment(&referenceCount) |> ignore
-        { new System.IDisposable with member x.Dispose() = this.DecrementUsageCount() }
+    /// Try to increment the usage count. If it returns None, then the object has been disposed by someone else already and may no longer be used.
+    member this.TryIncrementUsageCount() =
+        // Why is this so complex?
```
dsyme: Could you abstract all this out to a helper somewhere, e.g. in illib.fs please?   It shouldn't be hidden away here.  We might well need to reuse it.

Still doesn't seem it should be this complex, but I'll believe you :)

0x53A: Please DON'T believe me - I just happen to have read a few concurrency blog posts.

I have factored it out, but not yet tested it (using CI for that).

PLEASE DON'T PULL IT YET!

0x53A:
Please don't pull it yet, I'm not sure yet if the factored function correctly passes the byref through or if there is a value-copy hidden somewhere.

Using the CI as a smoke-test, and I want to do a second manual test ... tomorrow.

dsyme:
Closing as this is now well out of date

PR #729, opened 2017-03-27
Align with fsharp/fsharp & Microsoft/VisualFSharp via cherry-picking

cloudRoutine:
integrate with Microsoft/visualfsharp @ 91da9dbdf6a492372c0e3d497efcf0b726be357c


dsyme:
@cloudRoutine Hmm.... We normally do this via integrating (merging) 

    visualfsharp --> fsharp --> FSharp.Compiler.Service 

rather than cherry-picking.

cloudRoutine:
My mistake, the last time I had to move stuff across to them you told me to cherry-pick it, so I figured that was the way it was typically done.

I can split off all of the conifguration and organization commits and then reapply them on top of a merge if you think that'd be better

dsyme:
> I can split off all of the conifguration and organization commits and then reapply them on top of a merge if you think that'd be better

It's hard to tell.  Normally I just merge and fixup conflicts.  If it's equivalent to that then we can do that

cloudRoutine:
maybe i'll try again some other day...

dsyme:
@cloudRoutine I really, really want to get everything totally aligned and fully integrated.  We would then start building FSharp.Compiler.Service in Microsoft/visualfsharp.

It's possible we would still build the nuget package from here in case we need to make rapid fixes - but no feature development. The only feature drift I could imagine is Fable support, which requires a whole different test/tool chain to validate.

Perhaps we could make a big push on this alignment this week.

cloudRoutine:
i'd gotten to the point where it was building successfully and only 10 tests were failing, but I wasn't especially confident in all of my changes.

a suggestion on how we could make the alignment easier in the future - most places where `#if COMPILER` is used seem like types that it would be more useful to have the compiler service expose. So if we got rid of `COMPILER` as one of the defines for FCS add `|| COMPILER_SERVICE` to the 10 or so places where having `#if COMPILER` defined was actually useful and change all of the 

```fsharp
type (* internal *) foo = ...
```
to 
```fsharp
#if COMPILER
type internal foo 
#else
type foo
#endif 
    (args ) =
```
it should cut out a bit of merge conflict resolution busywork

dsyme:
@cloudRoutine Yeah, we have to just align that stuff somehow, agreed

PR #732, opened 2017-03-27
fix appveyor build

enricosada:
/cc @dsyme


enricosada:
@dsyme appveyor ok

dsyme:
Thank you!

PR #748, opened 2017-03-31
Deduplicate module names

alfonsogarciacaro:
Port of https://github.com/Microsoft/visualfsharp/pull/2728

Fixes https://github.com/Microsoft/visualfsharp/issues/2679

cc @dsyme @forki


dsyme:
Released as FCS 12.0.2

alfonsogarciacaro:
I updated the FCS dependency in Fable to 12.0.2 and I'm still having the same problem. @forki warned it was possible the fix didn't apply to FCS. I sent a specific test for FCS in #749. Sad.

PR #749, opened 2017-03-31
[WIP] Add test for #748

alfonsogarciacaro:
I cannot get the tests to run locally, neither for netcore or netfx, I see this:

```
Could not load file or assembly 'nunit.framework, Version=2.6.3.13283, Culture=neutral, PublicKeyToken=96d09a
1eb7f44a77' or one of its dependencies. The system cannot find the file specified.
```

Cannot run the new test in FSI either. I see this:
```
> ``add files with same name from different folders`` ()
- ;;
dllName = "test.dll", args = [|"--simpleresolution"; "--noframework"; "--debug:full"; "--define:DEBUG";
  "--optimize-"; "--out:test.dll"; "--doc:test.xml"; "--warn:3"; "--fullpaths";
  "--flaterrors"; "--target:library";
  "c:\Users\Alfonso\dev\FSharp.Compiler.Service\tests\service/data/samename/folder1/a.fs";
  "c:\Users\Alfonso\dev\FSharp.Compiler.Service\tests\service/data/samename/folder2/a.fs";
  "-r:C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.5\mscorlib.dll";
  "-r:C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.5\System.dll";
  "-r:C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.5\System.Core.dll";
  "-r:C:\Program Files (x86)\Reference Assemblies\Microsoft\FSharp\.NETFramework\v4.0\4.4.0.0\FSharp.Core.dll
"|]
System.IO.FileNotFoundException: Could not load file or assembly 'System.ValueTuple, Version=4.0.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified.
```

Let's see what CI says, then.

### DIFF
--- a/src/fsharp/CompileOps.fs
+++ b/src/fsharp/CompileOps.fs
@@ -3378,6 +3378,36 @@ let PostParseModuleSpecs (defaultNamespace,filename,isLastCompiland,ParsedSigFil
 
     ParsedInput.SigFile(ParsedSigFileInput(filename,qualName,scopedPragmas,hashDirectives,specs))
 
+
+/// Checks if a module name is already given and deduplicates the name if needed.
+let DeduplicateModuleName (moduleNamesDict:Dictionary<string,Set<string>>) (paths: Set<string>) path (qualifiedNameOfFile: QualifiedNameOfFile) =
+    let count = if paths.Contains path then paths.Count else paths.Count + 1
+    moduleNamesDict.[qualifiedNameOfFile.Text] <- Set.add path paths
+    let id = qualifiedNameOfFile.Id
+    if count = 1 then qualifiedNameOfFile else QualifiedNameOfFile(Ident(id.idText + "___" + count.ToString(),id.idRange))
+
+/// Checks if a ParsedInput is using a module name that was already given and deduplicates the name if needed.
+let DeduplicateParsedInputModuleName (moduleNamesDict:Dictionary<string,Set<string>>) input =
+    match input with
+    | ParsedInput.ImplFile (ParsedImplFileInput.ParsedImplFileInput(fileName,isScript,qualifiedNameOfFile,scopedPragmas,hashDirectives,modules,(isLastCompiland,isExe))) ->
+        let path = Path.GetDirectoryName fileName
+        match moduleNamesDict.TryGetValue qualifiedNameOfFile.Text with
+        | true, paths ->
+            let qualifiedNameOfFile = DeduplicateModuleName moduleNamesDict paths path qualifiedNameOfFile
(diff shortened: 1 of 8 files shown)


alfonsogarciacaro:
The test is failing :(

alfonsogarciacaro:
Wow, I can't believe this. Monkey patching worked first time! As @forki expected, FCS is not hitting the place where he put the fix for fcs.exe. After some navigation I copy & pasted the fix where it looked more plausible, but I'm not sure. @dsyme Could you please have a look? I don't know if I put the `seen` dictionary in the best place, it becomes a private member of `IncrementalBuilder` as I couldn't find a place to access all the `ParsedInput`s at once as in fsc.fs.

The best news is: it's working with Fable! Now I can go to sleep ;)

vasily-kirichenko:
A single (cached, may be discarded) instance of `IncrementalBuilder` is created for each project, so it looks like a right place for `seen` cache.

alfonsogarciacaro:
@vasily-kirichenko Thanks for the comment. I assume this also means that `IncrementalBuilder.ParseTask` won't be accessing `seen` from a different thread.

Now that you say that @forki, I realize that we may have the opposite problem: fsc doesn't hit the code in this PR. So we have two options:

- Keep the fix in the two places so it applies to both fsc and FCS. In that case it may be a good idea to convert the `deduplicate` function into a common helper (where?) parameterizing the `seen` dictionary.
- Try to move the code to `CompileOps.ParseOneInputFile` which seems to be common place to obtain the `ParsedInput`s both in https://github.com/Microsoft/visualfsharp/pull/2728 and this PR. In that case we probably need to make `seen` a parameter of `ParseOneInputFile` as this is a module helper and I guess it's not good to make it dependent of a static mutable value.

forki:
I vote for keeping it as is in both places for now without making it DRY.
Tbh I think I also need to fix it in fsi. Will do that in separate pr

Am 01.04.2017 13:13 schrieb "Alfonso Garcia-Caro" <notifications@github.com

forki:
So @dsyme please merge this one and @alfonsogcnunez please amend this to your VF# PR (so that VS tooling gets it as well). I will take care of fsi.

alfonsogarciacaro:
I have published a new Fable version with a custom FCS binary and it's working for now. We can put this on hold until thinking a better way to consolidate the fix among FCS, fsc and fsi 👍

Issue #658, opened 2016-10-22
FSharpSymbol do not carry information about enum fields


vasily-kirichenko:
For example, for `System.System.Text.RegularExpressions.RegexOptions`:

![image](https://cloud.githubusercontent.com/assets/873919/19619965/6872cd60-987b-11e6-9a16-fcfa627a0de8.png)

![image](https://cloud.githubusercontent.com/assets/873919/19619971/a073467c-987b-11e6-8eaa-1a2c10eeed31.png)

`FSharpFields` is used to create custom tooltips in MonoDevelop (and I'm trying to port it to VS Code at the moment) https://github.com/mono/monodevelop/blob/8b17d75ff2dc26c5aa1a6246f6a2ce7d4f65e3be/main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpSymbolHelper.fs#L681-L689

As `FSharpFields` are empty, both MonoDevelop and Code show the following tooltip:

![image](https://cloud.githubusercontent.com/assets/873919/19620006/50fc6ff0-987c-11e6-9eea-435827fd98d9.png)

![image](https://cloud.githubusercontent.com/assets/873919/19620008/66cc6e20-987c-11e6-90a1-da008e9ba864.png)

Currently we use `FSharpCheckFileResults.GetToolTipTextAlternate` https://github.com/fsharp/FSharp.Compiler.Service/blob/master/src/fsharp/vs/service.fs#L1909 and it does show enum fields properly:

![image](https://cloud.githubusercontent.com/assets/873919/19620017/8afb8664-987c-11e6-92ec-1cd77636a523.png)

dsyme:
Thanks, you're right this information should be provided

7sharp9:
This is specific to non F# enums is works fine for those defined in F#

![screen shot 2017-03-31 at 3 56 32 pm](https://cloud.githubusercontent.com/assets/588746/24555907/a5ffea8c-162a-11e7-9b31-8284150284fc.png)

7sharp9:
![screen shot 2017-04-03 at 2 45 10 pm](https://cloud.githubusercontent.com/assets/588746/24612179/66e4ce04-187c-11e7-885d-90aafc119ea2.png)

Found them.

7sharp9:
`ILTycon.Fields` also has them, I'll see if I can patch this up.

vasily-kirichenko:
I cannot reproduce in current VFT master:

[...]

PR #751, opened 2017-04-05
Enum fix

7sharp9:
Fix for #658 including test

### DIFF
--- a/src/fsharp/vs/Symbols.fs
+++ b/src/fsharp/vs/Symbols.fs
@@ -523,10 +523,22 @@ and FSharpEntity(cenv:cenv, entity:EntityRef) =
     member x.RecordFields = x.FSharpFields
     member x.FSharpFields =
         if isUnresolved() then makeReadOnlyCollection[] else
+    
+        if entity.IsILEnumTycon then
+            let (TILObjectReprData(scoref,enc,tdef)) = entity.ILTyconInfo
+            let formalTypars = entity.Typars(range.Zero)
+            let formalTypeInst = generalizeTypars formalTypars
+            let ty = TType_app(entity,formalTypeInst)
+            let formalTypeInfo = ILTypeInfo.FromType cenv.g ty
+            tdef.Fields.AsList
+            |> List.map (fun tdef -> let ilFieldInfo = ILFieldInfo(formalTypeInfo, tdef)
+                                     FSharpField(cenv, FSharpFieldData.ILField(cenv.g, ilFieldInfo) ))
+            |> makeReadOnlyCollection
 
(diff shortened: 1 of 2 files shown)


src/fsharp/vs/Symbols.fs
```diff
@@ -523,10 +523,22 @@ and FSharpEntity(cenv:cenv, entity:EntityRef) =
     member x.RecordFields = x.FSharpFields
     member x.FSharpFields =
         if isUnresolved() then makeReadOnlyCollection[] else
+    
+        if entity.IsILEnumTycon then
+            let (TILObjectReprData(scoref,enc,tdef)) = entity.ILTyconInfo
```
dsyme: I think these next few lines can be replaced by a use of ``generalizeTyconRef``?

src/fsharp/vs/Symbols.fs
```diff
@@ -523,10 +523,22 @@ and FSharpEntity(cenv:cenv, entity:EntityRef) =
...
+            let formalTypars = entity.Typars(range.Zero)
+            let formalTypeInst = generalizeTypars formalTypars
+            let ty = TType_app(entity,formalTypeInst)
+            let formalTypeInfo = ILTypeInfo.FromType cenv.g ty
+            tdef.Fields.AsList
```
dsyme: I think this can be done using GetILFieldInfosOfType on an infoReader, e.g. https://github.com/fsharp/FSharp.Compiler.Service/blob/0892b3c45ee4d3467a89e56b483919bb0271f5c2/src/fsharp/NicePrint.fs#L1491

(It's not too bad as is, but the uniform way is preferable)

tests/service/EditorTests.fs
```diff
@@ -564,6 +564,56 @@ let _ = arr.[..number2]
...
+    let allSymbols = typeCheckResults.GetAllUsesOfAllSymbolsInFile() |> Async.RunSynchronously
+    let enums =
+        allSymbols
+        |> Array.choose(fun s -> match s.Symbol with :? FSharpEntity as e when e.IsEnum -> Some e | _ -> None)
+        |> Array.distinct
+        |> Array.map(fun e -> (e.DisplayName, e.FSharpFields
+                                              |> Seq.map(fun f -> f.Name, f.LiteralValue )
+                                              |> Seq.toList))
+
+    enums |> shouldEqual
```
dsyme: great test, thank you

7sharp9:
@dsyme Ok, I'll take a look at the suggestions a bit later in the week.

dsyme:
I'll merge now, since the test case is fine and the cleanup is easy later

PR #758, opened 2017-04-19
Safer check for source code options in Project Cracker

cgravill:
The partitioning is looking for ".fs" as an extension but this can throw exceptions. This can happen on settings that aren't paths and therefore may contain invalid path characters e.g.  --ws:"Bundle" (a WebSharper 4.x setting).

### DIFF
--- a/src/fsharp/FSharp.Compiler.Service.ProjectCracker/ProjectCracker.fs
+++ b/src/fsharp/FSharp.Compiler.Service.ProjectCracker/ProjectCracker.fs
@@ -20,7 +20,7 @@ type ProjectCracker =
             let referencedProjects = Array.map (fun (a, b) -> a, convert b) opts.ReferencedProjectOptions
             
             let sourceFiles, otherOptions = 
-                opts.Options |> Array.partition (fun x -> x.IndexOfAny(Path.GetInvalidPathChars()) = -1 && Path.GetExtension(x).ToLower() = ".fs")
+                opts.Options |> Array.partition (fun x -> not (FileSystem.IsInvalidPathShim(x)) && Path.GetExtension(x).ToLower() = ".fs")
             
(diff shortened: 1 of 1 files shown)


src/fsharp/FSharp.Compiler.Service.ProjectCracker/ProjectCracker.fs
```diff
@@ -20,7 +20,7 @@ type ProjectCracker =
             let referencedProjects = Array.map (fun (a, b) -> a, convert b) opts.ReferencedProjectOptions
             
             let sourceFiles, otherOptions = 
-                opts.Options |> Array.partition (fun x -> Path.GetExtension(x).ToLower() = ".fs")
+                opts.Options |> Array.partition (fun x -> x.IndexOfAny(Path.GetInvalidPathChars()) = -1 && Path.GetExtension(x).ToLower() = ".fs")
```
dsyme: Looks like you can use ``not (FileSystem.IsInvalidPathShim(x))`` like this https://github.com/Microsoft/visualfsharp/blob/379f729dca9f310b474e3edf0669c71fe39b0129/src/fsharp/vs/IncrementalBuild.fs#L1612

dsyme: I fixed this with a direct edit to the PR

dsyme: Actually what you had was right - this code is not part of the compiler DLL itself

cgravill: Ah, that looks cleaner but that part of the code base seems quite independent. Thanks for the fast review.

PR #760, opened 2017-04-27
Update FAKE to 4.5.9

nosami:
This has a newer version of FCS and allows this solution to build with
Mono 5.0


dsyme:
@nosami Will all F# projects which use FAKE need that new version?

nosami:
Yeah or a later version .... there is a later fix that works around another bug by not attempting to cache the resulting dynamic assembly https://bugzilla.xamarin.com/show_bug.cgi?id=55681#c4

Issue #731, opened 2017-03-27
Failing tests on Mono 4.8 and Mono latest


dsyme:
Some tests have started to fail on Linux Mono latest:

Succeeds: https://travis-ci.org/fsharp/FSharp.Compiler.Service/builds/213030630

Failed: https://travis-ci.org/fsharp/FSharp.Compiler.Service/builds/213819074

These have identical mono versions 

    Mono JIT compiler version 4.8.0 (Stable 4.8.0.520/8f6d0f6 Wed Mar 15 16:18:35 UTC 2017)

But the F# package being used has changed:

```
Unpacking libfsharp-core4.3-cil (from .../libfsharp-core4.3-cil_4.0.0.3-0xamarin1_all.deb) ...
```
to
```
Unpacking libfsharp-core4.3-cil (from .../libfsharp-core4.3-cil_4.0.1.20-0xamarin3_all.deb) ...
```
With this new package the assembly FSharp.Core 4.3.0.0 is no longer being resolved. Also  the compiler used seems to have changed:

```
				Tool /usr/lib/cli/fsharp/fsc.exe execution started with arguments: ...
				Tool /usr/lib/mono/fsharp/fsc.exe execution started with arguments: ...
```

dsyme:
Here are the failures

```
Errors and Failures:
1) Test Failure : FSharp.Compiler.Service.Tests.ExprTests.Check use of type provider that provides calls to F# code
     Should not be errors, but: [|startup (1,1)-(1,1) typecheck error internal error: BuildFrameworkTcImports: no resolution of 'FSharp.Core.dll';
  unknown (1,1)-(1,1) typecheck error Unable to find the file 'FSharp.Core.dll' in any of
 /usr/lib/mono/4.5-api
 /home/travis/build/fsharp/FSharp.Compiler.Service/tests/service/data/TestProject
 /home/travis/build/fsharp/FSharp.Compiler.Service/bin/v4.5/|]
  Expected is <System.Object[0]>, actual is <Microsoft.FSharp.Compiler.FSharpErrorInfo[2]>
  Values differ at index [0]
  Extra:    < <startup (1,1)-(1,1) typecheck error internal error: BuildFrameworkTcImports: no resolution of 'FSharp.Core.dll'>, <unknown (1,1)-(1,1) typecheck error Unable to find the file 'FSharp.Core.dll' in any of
 /usr/lib/mono/4.5-api
[...]

PR #765, opened 2017-04-29
udpate to FSharp.Compiler.Tools 4.1.17

dsyme:
this should help on Mono latest, see #731


forki:
FAKE 4.61 should work

PR #786, opened 2017-06-28
docs --> docsrc

dsyme:
Goodbye gh-pages - put docs directly in master tree


forki:
o_O is that a good idea?

dsyme:
@forki I personally prefer it - but we can discuss

dsyme:
@forki Note this is the standard in ProjectScaffold now.

forki:
gh-pages is what github uses to view the page. For us it's only a deployment target. You write the docs still in markdown in the docs folder

forki:
> Note this is the standard in ProjectScaffold now.

o_O ?

dsyme:
See https://github.com/fsprojects/ProjectScaffold/blob/master/docsrc/tools/generate.template#L50. Documentation source is in "docsrc", output is in "docs", docs are published from master, see https://help.github.com/articles/configuring-a-publishing-source-for-github-pages/#publishing-your-github-pages-site-from-a-docs-folder-on-your-master-branch

forki:
why do you prefer it? we don't commit any artifacts? 
and I assume this can lead to pretty nasty merge conflicts

dsyme:
> why do you prefer it? we don't commit any artifacts? 

Probably because I've had repeated trouble keeping gh-pages up-to-date.  I've almost given up trying and the docs only get updated every 2-3 months as a result, or worse.  

I know "build ReleaseDocs" is meant to do the job, but it has consistenly failed late for me - right after I've tagged, pushed nuget packages, everything, and then pushing the docs fails because of some password problem or something random.  Then iterating on that is incredibly painful. With "docs",  it's just part of the PR, like AssemblyInfo

But you're right this may just be pushing the pain elsewhere

> and I assume this can lead to pretty nasty merge conflicts

Yes, possibly, you're right, especially for reference docs.  Will keep an eye on it and move back to the old model if it is causing diff churn.

forki:
`build ReleaseDocs` was there for that. It allowed to update the docs at any point in time

dsyme:
> build ReleaseDocs was there for that. It allowed to update the docs at any point in time

right, see above - it just kept failing late (during push) for too many obscure reasons I've also had a lot of trouble pushing to nuget.org from the build.  Nearly always related to credentials.  I hate credentials :)

forki:
`build ReleaseDocs` doesn't use any credentials. it's just a git push
I'm a bit worried about that change in ProjectScaffold tbh

dsyme:
> a bit worried about that change in ProjectScaffold tbh

Possibly it was a mistake.  Open a ProjectScaffold issue asking what users think?

forki:
started it here: https://github.com/fsprojects/ProjectScaffold/issues/304

Issue #788, opened 2017-07-08
Smallish typo in the Interactive Service Embedding page


mrakgr:
>Normally the one fromm FSharp.Compiler.Interactive.Settings.dll is used.

[The `fromm`](http://fsharp.github.io/FSharp.Compiler.Service/interactive.html#The-fsi-object) should be `from`.

Maybe it would be good to rewrite it as:

>Normally, the one from `FSharp.Compiler.Interactive.Settings.dll` is used.

Also...

> However, the F# interactive service is still useful, because you might want to wrap it in your own executable that is then executed (and communicates with the rest of your application), or if you only need to execute limited subset of F# code (e.g. generated by your own DSL).

The way the above sentence is phrased seems out of place. It feels like it is referring to something that is missing. Ordinarily, I'd fix the typo myself, but I am not sure what to do with the sentence above. Remove it?

dsyme:
Please propose a PR and we can discuss wording, thanks!

enricosada:
closing, fixed

PR #789, opened 2017-07-10
Fixed a typo.

mrakgr:


mrakgr:
This is related to issue #788.

Nothing really comes to mind as to what to do with the sentence I mentioned so please ignore that.

That having said I am trying to look up information on `--shadowcopyreferences` and cannot find it anywhere. It is not documented under FSI or FSC compiler options.

Issue #819, opened 2017-10-04
File name deduplication not working with ParseAndCheckFileInProject


alfonsogarciacaro:
The new `ParseAndCheckFileInProject` is working great to get the AST of a single file, thanks again @vasily-kirichenko for that! I'm currently adding some dependency checks to Fable to make sure all the necessary files are being refreshed when a watch compilation is triggered. I'm having some issues, but hopefully I should be able to fix them soon.

However, I've also noticed the trick implemented by @forki to deduplicate file names #748 (so you can have two files in different folders with the same name in the same project) doesn't work if I try to get the AST of a file with the same name as a previous one using `ParseAndCheckFileInProject` (it works if I use `ParseAndCheckProject`)

Any idea what may be the reason?

vasily-kirichenko:
All tools (VS, Ionide, Rider) use ParseAndCheckFileInProject, yet the latter two works nicely on Fable projects containing files with same name in different folders. So my guess is that you call the function in a wrong way. First question: do you pass full file name to it along with content?

dsyme:
@alfonsogarciacaro I think I'd need to see the actual calls, e.g. by submitting a failing test.  As @vasily-kirichenko says I suspect incorrect call parameters in some way

alfonsogarciacaro:
Sorry, I should have included instructions to reproduce the problem. You can do it with Fulma and Fable 1.3.0-beta:

```
git clone https://github.com/MangelMaxime/Fulma
cd Fulma
git checkout feature/test_fable_1_3
yarn
cd docs
dotnet restore
dontet fable webpack-dev-server
```

When Fable compilation is finished you can see the web in `localhost:8080`, then any change in any file will be automatically reflected. But if you edit the `src/Fulma/Common.fs` file you get the following error:

```
ERROR in ../src/Fulma/Common.fs
/Users/alfonsogarciacaronunez/dev/Fulma/src/Fulma/Common.fs(3,0): (75,0) error FSHARP: An implementation of the file or module 'Common' has already been given
 @ ./src/FulmaExtensions/PageLoader/View.fs 9:0-65
 @ ./src/FulmaExtensions/Dispatcher/View.fs
 @ ./src/App.fs
 @ ./docs.fsproj
 @ multi ../node_modules/webpack-dev-server/client?http://localhost:8080 webpack/hot/dev-server ./docs.fsproj
```

[...]

PR #820, opened 2017-10-05
 Fix de-duplication of module names in ParseAndCheckFileInProject (FCS bug #819)

dsyme:
Fix de-duplication of module names for ParseAndCheckFileInProject, see https://github.com/fsharp/FSharp.Compiler.Service/issues/819


dsyme:
This change is being integrated via #833

Issue #821, opened 2017-10-09
GetDeclarationLocation returns NoSourceCode if file does not exist.


amieres:
Right before finishing, GetDeclarationLocation checks to see if the file exists and if it does not then fails returning  DeclNotFound  NoSourceCode.
#### Repro steps

```
fsautocomplete
parse "test.fsx"
let abc = 1
let xyz = abc
<<EOF>>
{"Kind":"info","Data":"Background parsing started"}
{"Kind":"errors","Data":{"File":"<path>\\test.fsx","Errors":[]}}
finddecl "test.fsx" "let xyz = abc" 2 11
```

#### Expected behavior

```
Find declaration: ({Line = 2;
  Col = 11;}, 13, "let xyz = abc", ["abc"])
DeclFound
  <path>\test.fsx (1,4--1,7) IsSynthetic=false
{"Kind":"finddecl","Data":{"File":"<path>\\test.fsx","Line":1,"Column":5}}
```

#### Actual behavior

```
Find declaration: ({Line = 2;
  Col = 11;}, 13, "let xyz = abc", ["abc"])
DeclNotFound NoSourceCode
{"Kind":"error","Data":{"Code":1,"Message":"Could not find declaration","AdditionalData":{}}}
```
#### Known workarounds

Create an empty file "test.fsx"

#### Related information 

I understand this is on purpose, but is it necessary?
None of the other functionality like tooltips, complete, etc require the file to actually exist, 
why should this one?

While the workaroung is simple enough, it is a bit messy. 
Even worse though, is the process of finding out why it failed. Since FsAutoComplete does not return the reason "NoSourceCode" the only way to discover it is by debugging FsAutoComplete and then checking the source code for FSharpCompilerService to see why it didn't like it. 
NoSourceCode is a bit of a misnomer because the source code was already provided, it is the file that is missing not the source code.

dsyme:
@amieres I agree, please send a PR

amieres:
PR #822

PR #822, opened 2017-10-09
Removed check for NoSourceCode error from TypeCheckInfo.GetDeclarationLocation (issue #821)

amieres:
I did not remove NoSourceCode from type FSharpFindDeclFailureReason even though is probably not used anymore.


dsyme:
@amieres Do you think you could submit this to Microsoft/visualfsharp first?  Then we can integrate it here after it is accepted

amieres:
Ok here it is: 
https://github.com/Microsoft/visualfsharp/pull/3719

dsyme:
Thanks

PR #823, opened 2017-10-09
Make `IsRelativeNameResolvable` public

Krzysztof-Cieslak:
I've just hit it while trying to port SimpifyNameAnalyzer from VF# to FSAC (https://github.com/Microsoft/visualfsharp/blob/master/vsintegration/src/FSharp.Editor/Diagnostics/SimplifyNameDiagnosticAnalyzer.fs#L94)

### DIFF
--- a/src/fsharp/vs/service.fs
+++ b/src/fsharp/vs/service.fs
@@ -901,6 +901,18 @@ type TypeCheckInfo
                 Trace.TraceInformation(sprintf "FCS: recovering from error in IsRelativeNameResolvable: '%s'" msg)
                 false)
         
+    /// Determines if a long ident is resolvable at a specific point.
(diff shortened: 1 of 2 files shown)


Krzysztof-Cieslak:
OK, looks like it's not as simple - one of the parameters is of type `Compiler.NameResolution.Item`. It looks like whole `NameResolution` module is internal, and I guess it should stay like that.

I think we should change / add new override of `IsRelativeNameResolvable` that will take `SymbolUse` instead of `Item` and will do `symbolUse.Item` internally

dsyme:
@Krzysztof-Cieslak Yes, agreed

Could you submit to Microsoft/visualfsharp?  That makes things much easier for me, saves an RI

Krzysztof-Cieslak:
Sure, no problem

PR #826, opened 2017-10-18
Use paket in magic mode

forki:
Corresponding VF# PR https://github.com/Microsoft/visualfsharp/pull/3769


forki:
will come with next merge from VF# anyway

PR #882, opened 2018-12-07
Expose PrimaryAssembly parameter to AST compilation methods

eiriktsarpalis:
The current AST compilation facility does not expose a PrimaryAssembly parameter, meaning that targeting for netcore is currently failing.


baronfel:
@eiriktsarpalis should this be targeted upstream to visualfsharp/master as the canonical source? If so, I can integrate once it gets merged there.

cartermp:
This should be retargeted to here: https://github.com/dotnet/fsharp/

Issue #884, opened 2019-01-27
Support F# 4.6 with anonymous records


nightroman:
Is the latest FSharp.Compiler.Service (23.0.3) with FSharp.Core (4.6) supposed to work with anonymous records, or its newer version is needed? In my test project based on the above packages a script with anonymous records fails, such a syntax is not recognized.

#### Repro steps

- FSharp.Compiler.Service (23.0.3)
- FSharp.Core (4.6)
- Target framework is net45

Invoke some script with anonymous records.

#### Expected behavior

The script with anonymous records is compiler and invoked.

#### Actual behavior

The compilation fails due to not recognized syntax.

#### Known workarounds

None.

nojaf:
[25.0.1](https://www.nuget.org/packages/FSharp.Compiler.Service/25.0.1) is five months old, so I would assume FCS with support for anonymous records has not been released yet.

I think FCS releases go hand in hand with Visual Studio releases, however this is just speculation. Not sure.

Krzysztof-Cieslak:
FCS releases goes in hand with Don having time to sync repos and do release. :-)

If you need most recent version I’d recommend building it out of VF# repo, it’s fairly simple. That’s what I often do for FSAC/Ionide.

goswinr:
@Krzysztof-Cieslak is your more recent build from VF# for editor tooling also available as nuget ?

nightroman:
@Krzysztof-Cieslak Looking forward to anonymous records in Ionide!

baronfel:
@nightroman I'm waiting for permissions to push a few more related packages then I'll submit the MR for upgrades to FSAC and flow it through to Ionide.

PR #885, opened 2019-01-31
integrate visualfsharp's latest master

baronfel:
This seems to be ok, the tests and nuget seems to work alright, with one exception:

There are two tests that key off of the methods on `System.String`, but in some tests runs they include the new `AsMemory` and `AsSpan` methods, but some they do not.


baronfel:
This should help solve #884

dsyme:
Thank you!

dsyme:
Could you do an fcs version bump as well? https://github.com/fsharp/FSharp.Compiler.Service/blob/master/fcs/fcs.props#L6

dsyme:
Bump to 26.0.1 and search for 25.0.1 to check we get all occurrences

dsyme:
@baronfel Can you do the version bump in another PR? thanks :)

baronfel:
Yep, I'll get right on that. Thanks for the review/merge @dsyme!

PR #888, opened 2019-02-21
Merge in visualfsharp master, including better FCS support for Anonymous records

baronfel:
Covers from commits 5a5ca97 to b0292b8 (ie from January 31st to now)


dsyme:
Cool thanks!  Don't forget to bump the major version

PR #894, opened 2019-03-02
Integrate vfsharp feb 25 to mar 29

baronfel:
Integrate visualfsharp master from efb57cf56934bd28d3aa174d3c45a3eb4e7c1eb1 to af97b92e6.


baronfel:
I don't plan on merging this quite yet, I'm trying to package the nuget locally, flow it through into FSAC, then run tests against known-bad solutions and projects with ionide pointing to an updated FSAC instance to verify.  (Plus there are workflow things that need to be worked out in this repository to enable publishing of prerelease packages)

baronfel:
There's a fix identified and almost merged in VF#, I'll wait for that change to merge into upstream and then reintegrate before releasing a new version.

PR #896, opened 2019-03-04
Add symbol creation for an operator with a constraint

7sharp9:
This addresses a situaltion where if you request a tooltip for an operator with a constraint no symbol is returned due to:
```
Item.ImplicitOp(_, { contents = Some(TraitConstraintSln.FSMethSln(_, vref, _)) })
```
Being turned into a vanilla FSharpSymbol which has no real information about the symbol that the IDE can use.


7sharp9:
An corresponding issue was logged for ionide here: https://github.com/ionide/ionide-vscode-fsharp/issues/1025

goswinr:
thanks, shouldn't this go to VF# first / too ?
https://github.com/Microsoft/visualfsharp/blob/master/src/fsharp/symbols/Symbols.fs

7sharp9:
@goswinr It can, but I've always committed here first.

baronfel:
Closing this because the associated VF# issue was merged.

PR #910, opened 2019-09-23
Replace Microsoft/visualfsharp with dotnet/fsharp

Happypig375:
Replacing as much mentions of Microsoft/visualfsharp with dotnet/fsharp as possible.


baronfel:
Thank you!

PR #913, opened 2019-09-30
September integration done

baronfel:
This does the 4.7 migration and starts to fix documentation generation.


baronfel:
Not quite sure why the _number_ of commits reports like that, I only had to cherry-pick/rebase-onto the last 81 of them.

PR #917, opened 2019-11-26
Update downloaded dotnet sdk to version 3.0

eiriktsarpalis:
Attempt to fix the build


baronfel:
When the latest upstream changes are integrated we'll pick up 3.0.100, is that acceptable?

baronfel:
In addition to the new fake 5 runner, etc.

eiriktsarpalis:
That's fine. I posted this PR since the build was failing for me locally.

PR #926, opened 2020-02-05
Additional rebased commits on top of the january merge

baronfel:
Fixes runtime crashes on mono brought about by a few corner cases in mono's Memory-Mapped-File support and different runtime version numbers than .net framework.


baronfel:
With these commits merged/rebased fsharplint is able to update and run tests successfully, I'll be sending the proof PR over there as soon as this release is made.

PR #937, opened 2020-05-02
Update docs to FCS 35

nojaf:
@baronfel is this still the correct place for the docs?


baronfel:
yes, thank you for taking care of this.

PR #942, opened 2020-06-08
 Use Fornax for documentation generation

Krzysztof-Cieslak:


Krzysztof-Cieslak:
This is ready to review/test. Keep in mind it's not publishing docs - after generation (`build.sh GenerateDocs`) they're in `fcs\docsrc\_public`. 

My recommendation is using GH Actions to build and publish docs on new tag - https://github.com/SaturnFramework/Saturn/blob/master/.github/workflows/docs.yml

baronfel:
Thanks so much!

PR #949, opened 2020-06-16
Add JetBrains Rider to the project list

ForNeVeR:
I can see that this repo is supposed to be read-only, but it looks like it contains the actual FCS documentation, so I guess (?) this is the proper repository for this PR. Right?


baronfel:
Thanks :)

PR #969, opened 2021-11-05
Add FsDocsRepositoryBranch

nojaf:
Currently, the FCS API docs link to the master branch when inspecting the source link.
I believe this setting will correct this, based on https://github.com/fsprojects/FSharp.Formatting/blob/b62fe97ebcf7aa815480b5a2a535859ac5f5e081/src/FSharp.Formatting.Common/Templating.fs#L92-92
I haven't tested this to be honest 😅.


dsyme:
Thanks!

PR #971, opened 2022-08-30
strict build of docs

dsyme:


.github/workflows/docs.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     - name: Setup .NET Core
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 5.0.400
```
baronfel, line 21: instead of having to keep these versions up to date, if you have a global.json file in this repo you can update these actions to be:

```yaml
- name: Setup .NET Core
  uses: action/setup-dotnet@v2
```

and the action will use the version constraints from the global.json, keeping them in sync from that point onward.

dsyme: Looks like we need two sdks unfortunately, but it's all a bit messy.  FsDocs likes to run with .NET 6 and we're currently building docs from "main" of dotnet/fsahrp, which likes whatever .NET 7 preview it's on.

dsyme: I'm not even totally sure this separate docs build is worth it tbh, but if we want generated docs the only other solution would be to generate within dotnet/fsharp itself.  But that would probably lead to integrating all of FSharp.Formatting into dotnet/fsharp, which is a big hit, leading to more and more monorepo.  Perhaps it's what we should do.

dsyme:
Doc build is failing, because 

1. Project cracking needs needs updating to .NET 7
2. fsdocs needs needs updating to .NET 7
```
cracking projects...
  skipping project 'FSharp.Compiler.Service.fsproj' because an error occurred while cracking it: System.Exception: error - SDK Resolver Failure: "The SDK resolver "Microsoft.DotNet.MSBuildWorkloadSdkResolver" failed while attempting to resolve the SDK "Microsoft.NET.Sdk". Exception: "System.IO.FileNotFoundException: Could not load file or assembly 'System.Runtime, Version=7.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
File name: 'System.Runtime, Version=7.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
   at Microsoft.NET.Sdk.WorkloadMSBuildSdkResolver.WorkloadSdkResolver.Resolve(SdkReference sdkReference, SdkResolverContext resolverContext, SdkResultFactory factory)
   at Microsoft.Build.BackEnd.SdkResolution.SdkResolverService.ResolveSdk(Int32 submissionId, SdkReference sdk, LoggingContext loggingContext, ElementLocation sdkReferenceLocation, String solutionPath, String projectPath, Boolean interactive, Boolean isRunningInVisualStudio)""  D:\a\fsharp-compiler-docs\fsharp-compiler-docs\FSharp.Compiler.Service\FSharp.Compiler.Service.fsproj
log - 
   at Microsoft.FSharp.Core.PrintfModule.PrintFormatToStringThenFail@143[9](https://github.com/fsharp/fsharp-compiler-docs/runs/8093770493?check_suite_focus=true#step:9:10).Invoke(String message)
   at fsdocs.Crack.projectInfos@452.Invoke(String p) in /home/runner/work/FSharp.Formatting/FSharp.Formatting/src/fsdocs-tool/ProjectCracker.fs:line 454
Project cracking failed and --strict is on, exiting
```

baronfel:
There's a new version of proj-info with some fixes for 6.0.400 and onwards cracking that you'll probably need to take to unblock this, then.

dsyme:
@baronfel Updating some packages here https://github.com/fsprojects/FSharp.Formatting/pull/768

